PR #12, opened 2019-03-21
feat: list entries behind basic auth

nicosantangelo:
`/v1/entries`

### DIFF
--- a/src/S3/s3.ts
+++ b/src/S3/s3.ts
@@ -46,7 +46,7 @@ export async function listFiles(
   contents = contents.concat(data.Contents || [])
 
   return data.IsTruncated
-    ? listFiles(data.ContinuationToken, contents)
+    ? listFiles(data.NextContinuationToken, contents)
     : contents
 }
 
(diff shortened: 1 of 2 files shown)


src/S3/s3.ts
```diff
@@ -28,6 +28,28 @@ export function readFile(key: string): Promise<AWS.S3.GetObjectOutput> {
...
+  }
+
+  const listObjects = utils.promisify<AWS.S3.ListObjectsV2Output>(
+    s3.listObjectsV2.bind(s3)
+  )
+  const data = await listObjects(params)
+  contents = contents.concat(data.Contents || [])
+
+  return data.IsTruncated
+    ? listFiles(data.ContinuationToken, contents)
```
abarmat: The data returned by AWS has a field called `NextContinuationToken` instead of `ContinuationToken`. If changed to `NextContinuationToken` this function works!

nicosantangelo: but I'm using `listObjectsV2` which supposedly changed the name! that's also why the Typescript type `ListObjectsV2Output` works when doing `data.ContinuationToken`

nicosantangelo:
Yes! more info: https://docs.aws.amazon.com/cli/latest/reference/s3api/list-objects-v2.html

PR #14, opened 2019-03-31
feat: script to index scenes into db

abarmat:


scripts/scenes.sql
```diff
@@ -0,0 +1,20 @@
...
+  parcels_count INTEGER,
+  triangles_count INTEGER,
+  items_count INTEGER,
+  sessions INTEGER,
+  days INTEGER,
+  n_events INTEGER,
+  create_time INTEGER,
+  tag TEXT,
+  created_at TIMESTAMP,
+  updated_at TIMESTAMP
```
nicosantangelo: I think this looks different from the type https://github.com/decentraland/builder-contest-server/pull/14/files#diff-60e1ce117a51d7af5dd8d7072f3d7070R10

scripts/indexScenes.ts
```diff
@@ -0,0 +1,86 @@
+import * as fs from 'fs'
+import * as path from 'path'
+
+import * as crypto from '../src/crypto'
+import { Log, env } from 'decentraland-commons'
+import { db, Model } from 'decentraland-server'
+
+const log = new Log('IndexScenes')
+
+export interface OptionAttributes {
```
nicosantangelo: I think this should be called `SceneAttributes`?

scripts/indexScenes.ts
```diff
@@ -0,0 +1,86 @@
+import * as fs from 'fs'
+import * as path from 'path'
+
+import * as crypto from '../src/crypto'
```
nicosantangelo: If you want to be consistent with the import rules from the standards, these should look like:

```

import * as fs from 'fs'
import * as path from 'path'
import { Log, env } from 'decentraland-commons'
import { db, Model } from 'decentraland-server'

import * as crypto from '../src/crypto'
```

super minor tho

package.json
```diff
@@ -8,7 +8,8 @@
     "lint:fix": "tslint -p tsconfig.json",
     "build:tsc": "tsc -p tsconfig.json",
     "watch:build:tsc": "tsc -p tsconfig.json --watch",
-    "start": "npm run build:tsc && node dist/server.js"
+    "start": "npm run build:tsc && node dist/server.js",
+    "indexScenes": "tsc scripts/indexScenes.ts && node scripts/indexScenes.js"
```
nicosantangelo: `npm run build:tsc` already builds `scripts`:

```suggestion
    "indexScenes": "npm run build:tsc && node dist/scripts/indexScenes.js"
```

scripts/indexScenes.ts
```diff
@@ -0,0 +1,86 @@
...
+  return fs.readdirSync(baseDir)
+}
+
+const readScene = (filepath: string): any => {
+  const data = fs.readFileSync(filepath)
+  return JSON.parse(data.toString())
+}
+
+const connectDB = () => {
+  const CONNECTION_STRING = env.get('CONNECTION_STRING', undefined)
```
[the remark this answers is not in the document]
abarmat: If I remove the undefined default I'm getting
Argument of type 'string | {}' is not assignable to parameter of type 'string | undefined'.
  Type '{}' is not assignable to type 'string'.ts(2345)

Any ideas?

PR #16, opened 2019-05-31
feat: scene pool

abarmat:
Handles submissions of Projects for the Scene Pool.

### DIFF
--- a/src/Project/Project.router.ts
+++ b/src/Project/Project.router.ts
@@ -3,7 +3,7 @@ import { server } from 'decentraland-server'
 
 import { Router } from '../common'
 import { encrypt, decrypt } from '../crypto'
-import { readEntry, saveEntry, getUploader, EntryPrefix } from '../storage'
+import { readEntry, saveEntry, getFileUploader, EntryPrefix } from '../storage'
 import { Entry } from './types'
 import { parseEntry } from './validations'
 
@@ -19,8 +19,7 @@ export class ProjectRouter extends Router {
      */
     this.router.post(
       '/project/:projectId/preview',
-      server.handleRequest(this.submitPreview),
-      getUploader().array('', 2)
+      server.handleRequest(this.submitPreview)
     )
   }
 
@@ -49,11 +48,25 @@ export class ProjectRouter extends Router {
     return true
   }
 
-  async submitPreview(req: express.Request): Promise<boolean> {
+  async submitPreview(
+    req: express.Request,
+    res: express.Response
+  ): Promise<boolean> {
     const projectId = server.extractFromReq(req, 'projectId')
-    console.log(projectId)
-    // Check if project id exists
 
-    return true
+    // Check if project id exists
+    const entry: Entry = await readEntry(projectId, EntryPrefix.Project)
+    if (!entry) {
+      throw new Error('Cannot add files to non-existing project')
+    }
+
+    // TODO: add file extension based on mime/type
+    return getFileUploader(projectId, EntryPrefix.Project).single('attachment')(
+      req,
+      res,
+      () => {
+        return true
+      }
+    )
   }
 }
--- a/src/Project/Project.router.ts
+++ b/src/Project/Project.router.ts
@@ -43,9 +43,9 @@ export class ProjectRouter extends Router {
   }
 
   async submitProject(req: express.Request): Promise<boolean> {
-    const EntryJSON = server.extractFromReq(req, 'entry')
+    const entryJSON = server.extractFromReq(req, 'entry')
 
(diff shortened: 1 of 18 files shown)


src/Project/Project.router.ts
```diff
@@ -0,0 +1,115 @@
...
+
+import { Router } from '../common'
+import { encrypt, decrypt } from '../crypto'
+import { checkFile } from '../S3'
+import { readEntry, saveEntry, getFileUploader, EntryPrefix } from '../storage'
+import { Entry } from './types'
+import { parseEntry } from './validations'
+
+const SUPPORTED_FILE_FIELDS = [
+  'image',
```
belohlavek: `image` is not required

src/storage/index.ts
```diff
@@ -0,0 +1,79 @@
...
+  bucketName,
+  checkFile,
+  parseFileBody,
+  readFile,
+  s3,
+  uploadFile
+} from '../S3'
+
+const ENTRY_FILENAME = 'entry.json'
+const MAX_FILE_SIZE = 3000000
```
belohlavek: Let's make it 5mb, this will apply to the JSON as well

src/storage/index.ts
```diff
@@ -0,0 +1,79 @@
...
+    const file = await readFile(key)
+    return parseFileBody(file)
+  } catch (error) {
+    // No previous entity
+  }
+}
+
+export async function saveEntry(
+  projectId: string,
+  entry: any,
```
belohlavek: can't this be of type Entry?

src/storage/index.ts
abarmat: The limit for the JSON is set at a different level:
`this.app.use(bodyParser.json({ limit: '4mb' }))` 

I will change that one to `5mb`.

src/Project/Project.router.ts
abarmat: 👍 Removed

src/storage/index.ts
abarmat: This function serves both for ProjectEntry and ContestEntry they currently don't share a common parent type. I will make it inherit from a Base type so we can enforce the type validation.  👍

src/Project/Project.router.ts
```diff
@@ -0,0 +1,117 @@
...
+
+    return next()
+  }
+
+  async filesUploaded(req: express.Request, _res: express.Response) {
+    const uploadedFiles = Object.values(req.files)
+
+    // Check if all files uploaded
+    const areFilesUploaded = uploadedFiles
+      .map(files => {
```
nicosantangelo: I think this could be a single `every` call with the same return value. Changing `map` to `every` and removing the `every` below should work

src/storage/index.ts
```diff
@@ -0,0 +1,82 @@
...
+  entry: ContestEntry | ProjectEntry,
+  prefix: EntryPrefix
+) {
+  const key = getEntryKey(projectId, prefix)
+  await uploadFile(key, Buffer.from(JSON.stringify(entry)))
+  await checkFile(key)
+}
+
+export function getFileUploader(prefix: EntryPrefix, acl: string = 'private') {
+  return multer.default({
```
nicosantangelo: You may want to add a [`fileFilter`](https://github.com/expressjs/multer#filefilter) function here, to allow only certain file types:

```ts
// Example code, don't copy paste :D
    fileFilter(_, file, cb) {
      cb(null, Object.keys(contentTypeExtensionMap).includes(file.mimetype))
    }
```

src/storage/index.ts
```diff
@@ -0,0 +1,82 @@
...
+      bucket: bucketName,
+      key: function(req: express.Request, file, cb) {
+        const fileExtension = contentTypeExtensionMap[file.mimetype]
+        if (!fileExtension) {
+          cb(new Error(`Content-Type not supported : ${file.mimetype}`))
+          return
+        }
+
+        const projectId = server.extractFromReq(req, 'projectId')
+        const key = `${prefix}/${projectId}/${file.fieldname}.${fileExtension}`
```
nicosantangelo: it feels like this should be build by the same function that does `getEntryKey`. They're generating the same key (as far as I can tell) and changing one should change the other, no?

src/Project/Project.router.ts
abarmat: 👍

src/storage/index.ts
abarmat: 👍

src/storage/index.ts
abarmat: Will do

PR #26, opened 2019-07-25
feat: add database migrations

nicosantangelo:
```
npm run migrate create {migration-name} - creates a new migration file with the name you give it. Spaces and underscores will be replaced by dashes and a timestamp is prepended to your file name.
npm run migrate up - runs all up migrations from the current state.
npm run migrate up {N} - runs N up migrations from the current state.
npm run migrate down - runs a single down migration.
npm run migrate down {N} - runs N down migrations from the current state.
npm run migrate unlock - unlocks migrations (if previous up/down migration failed and was not automatically unlocked).
npm run migrate redo - redoes last migration (runs a single down migration, then single up migration).
npm run migrate redo {N} - redoes N last migrations (runs N down migrations, then N up migrations).
```

From https://github.com/salsita/node-pg-migrate/tree/v2#usage
You'll need to add a `.env` on the root of the project. See `.env.example`

### DIFF
--- a/migrations/migrate.ts
+++ b/migrations/migrate.ts
@@ -1,0 +1,59 @@
+#!/usr/bin/env ts-node
+
+import * as path from 'path'
+import { spawn } from 'child_process'
+import { env } from 'decentraland-commons'
+
+export function migrate(
+  commandArguments: string[],
+  migrationsDir: string = __dirname
+) {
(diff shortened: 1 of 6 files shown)


cazala:
Approved but needs conflicts to be resolved

cc @nicosantangelo

nicosantangelo:
@cazala Updated :)

PR #27, opened 2019-07-25
feat: project crud

nicosantangelo:
**Test with postman**
[Decentraland - Builder server.postman_collection.json.zip](https://github.com/decentraland/builder-server/files/3449236/Decentraland.-.Builder.server.postman_collection.json.zip)

[media.zip](https://github.com/decentraland/builder-server/files/3449235/media.zip)

**Endpoints**
- GET /projects
- GET /projects/:id
- PUT /projects/:id
- DELETE /projects/:id
- PUT /projects/:id/manifest
- DELETE /projects/:id/manifest
- POST /projects/:id/media

**Run**

```bash
createdb builder
cp .env.example .env
vim .env # CONNECTION_STRING='postgres://localhost:5432/builder'

npm install
npm run migrate up

npm start
```

Closes https://github.com/decentraland/builder/issues/560

PR #28, opened 2019-07-31
feat: deployment crud

nicosantangelo:
**Depends** on https://github.com/decentraland/builder-server/pull/27

**Test with postman**
[Decentraland - Builder server - Deployments.postman_collection.json.zip](https://github.com/decentraland/builder-server/files/3466814/Decentraland.-.Builder.server.-.Deployments.postman_collection.json.zip)

**Endpoints**
- GET /deployments
- GET /projects/:id/deployment
- PUT /projects/:id/deployment
- DELETE /projects/:id/deployment

**Run**

```bash
npm run migrate up
npm start
```


migrations/1564534118044_create-deployments.ts
```diff
@@ -0,0 +1,28 @@
...
+const tableName = Deployment.tableName
+
+export const up = (pgm: MigrationBuilder) => {
+  pgm.createTable(
+    tableName,
+    {
+      id: { type: 'UUID', primaryKey: true, unique: true, notNull: true },
+      user_id: { type: 'TEXT', notNull: true },
+      lastPublishedCID: { type: 'TEXT' },
+      isDirty: { type: 'BOOLEAN', notNull: true },
```
cazala: should we name `lastPublishedCID` and `isDirty` also in snake_case? We are already doing a transformation from the response in the frontend so we could transform those too and keep the naming consistent here.

nicosantangelo: Indeed! I missed those two and used the old test deployment so I didn't realised.

PR Updated, both the code and Postman test files

Issue #31, opened 2019-08-05
Pool endpoint


cazala:
We need and endpoint to add a users scene to the LAND pool. This endpoint should store a snapshot of the project data, the manifest, and the media associated with that project.

A possible implementation could be:

```
POST /pool
Authorization: Bearer {token}
{ "id": "77baa481-611e-460d-9642-1bb4d7ed659e" }
```

That upserts a row in a `Pool` table that has the same schema as the `Project` table, copies the project manifest in s3 to a `pool` directory, along with the media for that project (thumbnail and screenshots). This endpoint should be protected and return a 401 if the user is unauthenticated or unauthorized.

```
GET /pool
{ "ok" true, "data": [ array of projects ] }
```

Returns all the projects in the pool. Once we do pagination and filters for the `/projects` endpoint we should apply those too to this one.

nicosantangelo:
Remember add `/pools/:id/manifest`

PR #33, opened 2019-08-06
feat: sort and paginate models

nicosantangelo:
**Postman**

[Filter collection.postman_collection.json.zip](https://github.com/decentraland/builder-server/files/3476685/Filter.collection.postman_collection.json.zip)

**ENDPOINTS**

- GET `/projects?limit=1&offset=2&sort_by=created_at&sort_order=ASC`

Adds generic **modifiers** for the query string:
- `limit`: amount of items to return
- `offset`: how many items should be skipped
- `sort_by`: one of the whitelisted properties of the Model
- `sort_order`: ASC or DESC
- `XXX_eq`: XXX should be the name of a whitelisted property, for example: `title_eq=Hi`
- `XXX_not_eq`: XXX should be the name of a whitelisted property, for example: `description_not_eq=Bye`

**Whitelisted Properties**
- title
- description
- cols
- rows
- created_at
- updated_at

Closes https://github.com/decentraland/builder-server/issues/32

PR #34, opened 2019-08-08
feat: pools

nicosantangelo:
**Postman**
[Builder Server - Pool.postman_collection.json.zip](https://github.com/decentraland/builder-server/files/3481526/Builder.Server.-.Pool.postman_collection.json.zip)

**ENDPOINTS**

- GET `/pools?limit=1&offset=2&sort_by=created_at&sort_order=ASC`
- GET `/projects/:id/pool
- GET `/pools/:id/manifest
- PUT `/project/:id/pool

Continues #33 
Closes: https://github.com/decentraland/builder-server/issues/31


src/Project/Project.router.ts
```diff
@@ -126,6 +131,10 @@ export class ProjectRouter extends Router {
       throw new HTTPError('Invalid schema', validator.errors)
     }
 
+    if (!(await Project.canUpsert(id, user_id))) {
```
cazala: Why are we using this instead of the `projectAuthorization` middleware?

nicosantangelo: Because projectAuthorizations does the following:
`await Project.isOwnedBy(id, user_id)`

but when you're upserting the project, the first insert will never work!
That's why we do this check:

```ts
static async canUpsert(id: string, userId: string) {
  const [projectExists, isOwner] = await Promise.all([
    Project.exists(id),
    Project.isOwnedBy(id, userId)
  ])
  return !projectExists || isOwner
}
```

you can upsert if the project does not exist (first time) or if you're the owner. The same thing happens on the manifest upsert

Does that make sense?

cazala: Totally :+1:

PR #38, opened 2019-08-09
feat: get media

nicosantangelo:
**TO RUN**
```bash
npm install
```

**ENDPOINTS**

- GET `/projects/8673ea12-3137-433f-9206-d84795982487/media/preview.png`

**DEPENDS ON**

#34


cazala:
I'm getting the following when I `GET /projects/:id/manifest`:

```{"ok":false,"data":{},"error":"Unexpected token o in JSON at position 1"}```

It doesn't happen when I do the same request with the same DB on the `master` branch.

cazala:
If it helps, it doesn't happen on the `feat/pools` branch either, so it looks like it's something related to the changes introduced in this PR

nicosantangelo:
Good catch! had to extract some logic and there was a missing `await`

PR #74, opened 2019-10-15
feat: social share

2fd:


migrations/1571079562023_add-is-public-to-projects.ts
```diff
@@ -0,0 +1,20 @@
...
+    notNull: true
+  }
+}
+
+export const up = (pgm: MigrationBuilder) => {
+  pgm.addColumn(tableName, columns)
+}
+
+export const down = (pgm: MigrationBuilder) => {
+  pgm.dropColumn(tableName, columnName)
```
nicosantangelo: This is OK but in case you didn't know, the `down` method can be safely inferred by pgm using the `up` command (meaning that it's not necessary)

src/Manifest/Manifest.router.ts
```diff
@@ -68,6 +73,14 @@ export class ManifestRouter extends Router {
     }
   }
 
+  async getPublicProjectManifest(req: AuthRequest) {
```
nicosantangelo: Is this method necessary? can't we just use `getProjectManifest`? (the `public` aspect of this seems to be the lack of middlewares)

src/Pool/Pool.router.ts
```diff
@@ -31,13 +31,7 @@ export class PoolRouter extends Router {
     /**
      * Get pool
      */
-    this.router.get(
-      '/projects/:id/pool',
-      withAuthentication,
-      withProjectExists,
```
nicosantangelo: Is this one not necessary anymore? `withProjectExists`
I'm not sure if it's a requirement for a pool to have a valid (active) project, we could ask

src/Project/Project.router.ts
```diff
@@ -84,6 +84,11 @@ export class ProjectRouter extends Router {
       server.handleRequest(this.deleteProject)
     )
 
+    this.router.get(
+      '/projects/:id/public',
```
nicosantangelo: I think we can maybe keep `withProjectExists` here, to be at least more consistent with the other endpoints.

If not, maybe it's a good idea to check the existence of `Project.findOne({ id, is_public: true })` (below) before returning it, so we can return a `404` like the other endpoints.

src/Project/Project.types.ts
```diff
@@ -9,6 +9,7 @@ export type ProjectAttributes = {
   rows: number
   created_at: Date
   updated_at: Date
+  is_public: boolean
```
nicosantangelo: nitpicking here, but can we move this above `created_at`? to keep it consistent. Same goes with `projectSchema`

migrations/1571079562023_add-is-public-to-projects.ts
2fd: 👍

src/Pool/Pool.router.ts
2fd: on the frontend side, isn't a requirement.
 @cazala @belohlavek ?

PR #75, opened 2019-10-25
feat: Social share

2fd:
- fix: publish to pool
- feat: add share redirections
- feat: add environment variable `BUILDER_URL`:

| environment | value                               |
|-------------|-------------------------------------|
| local       | `BUILDER_URL=http://localhost:3000`             |
| stg         | `BUILDER_URL=https://builder.decentraland.zone` |
| prd         | `BUILDER_URL=https://builder.decentraland.org`  |


migrations/1572013310034_add-is-public-to-pool.ts
```diff
@@ -0,0 +1,16 @@
+import { MigrationBuilder } from 'node-pg-migrate'
+import { Pool } from '../src/Pool'
+
+const tableName = Pool.tableName
+const columnName = 'is_public'
+const columns = {
+  [columnName]: {
+    type: 'BOOLEAN',
+    default: true,
```
nicosantangelo: Wouldn't it be better to just not this prop here and treat every pool like it's public?
Mostly because we're already doing this:
```
Pool.findOne<PoolAttributes>({ id })
```
here: https://github.com/decentraland/builder-server/pull/75/files#diff-64684767296fe331467d8529f62c5dacR58

Below we should do something like new Pool(utils.omit(project!, ['is_public'])).upsert() if we do go ahead with this option

src/Share/Share.router.ts
```diff
@@ -0,0 +1,63 @@
...
+    const p = await this.findPoolOrProject(id, type)
+
+    if (!p) {
+      return res.send(404)
+    }
+
+    const thumbnail =
+      p.thumbnail &&
+      format(
+        `http://localhost:5000/v1/projects/%s/media/thumbnail.png?updated_at=%s`,
```
nicosantangelo: is this correct? shouldn't this be using another URL?

nicosantangelo: Also, what's the advantage of using `util`'s format instead of just using a template string?

```
`http://localhost:5000/v1/projects/${p.id}/media/thumbnail.png?updated_at=${Date.parse(p.updated_at.toString())}\`
```

src/Share/Share.router.ts
```diff
@@ -0,0 +1,63 @@
...
+        p.id,
+        Date.parse(String(p.updated_at))
+      )
+
+    return res.status(200).send(template({ ...p, url, thumbnail }))
+  }
+
+  private async findPoolOrProject(
```
nicosantangelo: Maybe we can leave out the actual thing we're looking for here and delegate that responsibility to the `type` argument. So maybe this method could be called

`findByType` or `findElementByType`?

What do you think?

src/Share/Share.router.ts
```diff
@@ -0,0 +1,63 @@
...
+      '/share/:type(scene|pool)/:id',
+      withSocialUserAgentDetector,
+      this.redirectToBuilder
+    )
+  }
+
+  private redirectToBuilder = async (req: SocialRequest, res: Response) => {
+    const { type, id } = req.params as Params
+    const targetPath = type === 'scene' ? `/view/${id}` : `/view/${type}/${id}`
+    const url = BUILDER_URL + targetPath
```
nicosantangelo: This scares me _a bit_. Maybe we should use `url.resolve` to avoid messing up here?

src/Share/Share.router.ts
```diff
@@ -0,0 +1,63 @@
...
+    if (!req.socialAgent) {
+      return res.redirect(301, url)
+    }
+
+    const p = await this.findPoolOrProject(id, type)
```
nicosantangelo: Same as below, maybe naming it `p` is shooting ourselves in the foot? What if tomorrow we add a new element that starts with a different letter.
Do we maybe have a name that encompasses both project and pools?

src/Share/Share.router.ts
```diff
@@ -0,0 +1,63 @@
...
+      )
+
+    return res.status(200).send(template({ ...p, url, thumbnail }))
+  }
+
+  private async findPoolOrProject(
+    id: string,
+    type: 'pool' | 'scene'
+  ): Promise<ProjectAttributes | PoolAttributes | undefined> {
+    if (type === 'pool') {
```
nicosantangelo: **Suggestion**

You think it'd be beneficiary to add a switch and check for every option? throwing in case the type is not recognized:

```ts
switch(type) {
  case 'pool':
    return return Pool.findOne<PoolAttributes>({ id })
  case 'scene':
    return Project.findOne<ProjectAttributes>({ id, is_public: true })
  default:
    throw new Error(`Unknown type ${type} for id ${id}`)
}
```

I know typescript is enforcing the type here, but we _could_ use `as any` at the call site.

nicosantangelo: Furthermore, you think it'd be a good idea to have a `enum ElementType` (using element as a concept here, but name pending)?
I first thought that it might local to extract, but we're repeating this on the definition of the `Param` type on `Share.types.ts`

src/middleware/share.ts
```diff
@@ -0,0 +1,164 @@
+import { Request, Response, NextFunction } from 'express'
+// import { server } from 'decentraland-server'
+import { env } from 'decentraland-commons'
+
+const BUILDER_DOMAIN = env.get('BUILDER_DOMAIN')
+if (!BUILDER_DOMAIN) {
```
nicosantangelo: I think this should be `BUILDER_URL`? or maybe it's missing from `.env.example`?

src/middleware/share.ts
```diff
@@ -0,0 +1,164 @@
...
+    // ]
+  },
+
+  {
+    name: 'pocket',
+    pattern: 'PocketParser'
+    // url: 'https://getpocket.com/pocketparser_ua',
+    // instances: [
+    //   'PocketParser/2.0 (+https://getpocket.com/pocketparser_ua)'
+    // ]
```
nicosantangelo: I don't know what I'm talking about here, but are the comments for each of these necessary?

src/Share/Share.router.ts
```diff
@@ -0,0 +1,63 @@
+import { Response } from 'express'
+import { env } from 'decentraland-commons'
+import { format } from 'util'
```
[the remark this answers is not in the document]
2fd: is a NodeJs module: https://nodejs.org/dist/latest-v12.x/docs/api/util.html

src/Share/Share.router.ts
2fd: It's the same but looks less wierd than:

Screen Shot 2019-10-29 at 10 33 07

> ☝️   code after format

src/middleware/share.ts
2fd: Yes, I don't want to lose these references

2fd:
@nicosantangelo 👍

nicosantangelo:
The code LGTM, I'm testing the endpoints !

PR #90, opened 2019-11-29
feat: add pool like

2fd:


src/PoolLike/PoolLike.types.ts
```diff
@@ -0,0 +1,22 @@
+export type PoolLikeAttributes = {
+  pool: string
+  user: string
+  created_at: string
+}
+
+export const poolSchema = {
```
nicosantangelo: I think this should be `poolLikeSchema`

src/PoolLike/PoolLike.router.ts
```diff
@@ -0,0 +1,81 @@
...
+
+    const exists = await PoolLike.count({ pool, user })
+
+    if (!exists) {
+      await PoolLike.create({ pool, user, created_at: new Date() })
+      const likes = await PoolLike.count({ pool })
+      await Pool.update({ likes }, { id: pool })
+    }
+
+    return true
```
nicosantangelo: Wouldn't it be useful/necessary to return a more comprehensive response? for example returning the PoolLike record, or maybe the total count?

That way the consumer could see if the like already existed for example

src/PoolLike/PoolLike.router.ts
```diff
@@ -0,0 +1,81 @@
...
+
+    const exists = await PoolLike.count({ pool, user })
+
+    if (exists) {
+      await PoolLike.delete({ pool, user })
+      const likes = await PoolLike.count({ pool })
+      await Pool.update({ likes }, { id: pool })
+    }
+
+    return true
```
nicosantangelo: Same as above! maybe it's useful to return a more comprehensive response?

src/Pool/Pool.router.ts
```diff
@@ -78,11 +92,23 @@ export class PoolRouter extends Router {
   }
 
   async getPool(req: AuthRequest) {
-    const id = server.extractFromReq(req, 'id')
-    return Pool.findOne({ id })
+    const pool_id = server.extractFromReq(req, 'id')
+    const user_id = (req.auth && req.auth.sub) || null
```
nicosantangelo: I think we're doing this: `requestParameters.get('user_id', null)` on `getPools`. Maybe we can do the same here?

Also, as a TODO (not now), we might want to have a PermissiveAuthRequest or something of the sort for these casese

src/PoolLike/PoolLike.types.ts
```diff
@@ -0,0 +1,23 @@
...
+  user_id: string
+  created_at: string
+}
+
+export type PoolLikeCount = {
+  pool_id: string
+  user_id?: string
+}
+
+export const poolLikeSchema = {
```
nicosantangelo: Should we delete `poolLikeSchema`? it's not really being used. Or should we use it in the corresponding endpoint?

src/Pool/Pool.router.ts
2fd: This expression is for determining if the current user (if there is) gives like to the pool. You don't want to load a pool with the like of an arbitrary user

PR #113, opened 2020-08-20
feat: add collections and items

nicosantangelo:
```
GET /items
GET /collections
GET /collections/:id
GET /collections/:id/items

PUT /items/:id
PUT /collections/:id

POST /items/:id/files
```


migrations/1597864779562_create-items.ts
```diff
@@ -0,0 +1,33 @@
...
+const tableName = Item.tableName
+
+export const up = (pgm: MigrationBuilder) => {
+  pgm.createTable(
+    tableName,
+    {
+      id: { type: 'UUID', primaryKey: true, unique: true, notNull: true },
+      name: { type: 'TEXT', notNull: true },
+      description: { type: 'TEXT' },
+      eth_address: { type: 'TEXT', notNull: true },
```
nachomazzara, line 13: Maybe `eth_address` can be interpreted to the Ethereum address of the item, What do you think of using `owner` or `creator`? We know that in the future we will need to have off/on-chain permissions, so we need to use a name that won't hurt.

migrations/1597864779562_create-items.ts
```diff
@@ -0,0 +1,33 @@
...
+      collection_id: { type: 'UUID' },
+      blockchain_item_id: { type: 'TEXT' },
+      price: { type: 'TEXT' },
+      beneficiary: { type: 'TEXT' },
+      rarity: { type: 'TEXT' },
+      type: { type: 'TEXT', notNull: true },
+      data: { type: 'JSON', notNull: true },
```
nachomazzara, line 20: QQ: Are we going to use `data` as a JSON then? I thought that we will use `data_id` creating different tables for each type of data.

migrations/1597882126530_create-collections.ts
```diff
@@ -0,0 +1,27 @@
...
+const tableName = Collection.tableName
+
+export const up = (pgm: MigrationBuilder) => {
+  pgm.createTable(
+    tableName,
+    {
+      id: { type: 'UUID', primaryKey: true, unique: true, notNull: true },
+      name: { type: 'TEXT', notNull: true },
+      eth_address: { type: 'TEXT', notNull: true },
+      hash: { type: 'UUID' },
```
nachomazzara, line 13: I'd suggest using  name `salt` and type `TEXT` which is better to store a 32 bytes varchar. `hash` is more generic. In some languages is a reserved word.

migrations/1597882126530_create-collections.ts
```diff
@@ -0,0 +1,27 @@
...
+      contract_address: { type: 'TEXT' },
+      is_published: { type: 'TEXT', default: false, notNull: true },
```
nachomazzara, line 15: What do you think of using `BOOLEAN` here?

src/Collection/Collection.router.ts
```diff
@@ -0,0 +1,92 @@
...
+
+  async getCollections(req: AuthRequest) {
+    const eth_address = req.auth.ethAddress
+    return Collection.find({ eth_address })
+  }
+
+  async getCollection(req: AuthRequest) {
+    const id = server.extractFromReq(req, 'id')
+    const eth_address = req.auth.ethAddress
+    return Collection.findOne({ id, eth_address })
```
nachomazzara, line 59: Is `eth_address` useful here?

src/Item/Item.router.ts
```diff
@@ -0,0 +1,120 @@
...
+
+  async getItems(req: AuthRequest) {
+    const eth_address = req.auth.ethAddress
+    return Item.find({ eth_address })
+  }
+
+  async getCollectionItems(req: AuthRequest) {
+    const id = server.extractFromReq(req, 'id')
+    const eth_address = req.auth.ethAddress
+    return Item.find({ eth_address, collection_id: id })
```
nachomazzara, line 71: Same question here related to `eth_address`

migrations/1597864779562_create-items.ts
nicosantangelo: The problem is that AssetPacks, Assets, etc use `eth_address` to represent the user's address. I thought of owner as well when we where setting up the database but juanca reminded us of that. Furthermore, parts of the code rely on that property being called `eth_address` to work generically. Like Ownable

migrations/1597882126530_create-collections.ts
nicosantangelo: Gotcha!
The UUID was a slipup btw

migrations/1597882126530_create-collections.ts
nicosantangelo: It should be BOOLEAN, good catch

src/Collection/Collection.router.ts
nicosantangelo: I thought about not having it here to be honest, but I left it as a check. But now that I think about it that's being handled on the middleware so no need. I'll remove it !

cazala:
Should we add the new env vars `FACTORY_COLLECTION_ADDRESS` and `FACTORY_COLLECTION_CODE_HASH` to the `.ci` config? 

https://github.com/decentraland/builder-server/blob/master/.ci/index.ts

nicosantangelo:
Yes! we don't have either value yet but I can add them as an empty string for now

nachomazzara:
I think that would be better to have addresses and the code hash inside the code in constant dictionaries to work with different networks without changing every time the env.

```typescript
export const FACTORY_ADDRESSES = {
1: '0x..123',
3: '0x...234'
}

export const FACTORY_CREATION_CODE_HASH = {
1: '0x123...',
3: '0x455...'
}
```

PR #118, opened 2020-09-02
fix: schemas & upload

cazala:


src/Asset/Asset.router.ts
```diff
@@ -73,9 +73,11 @@ export class AssetRouter extends Router {
   }
 
   private getAssetFilesRequestHandler() {
-    const uploader = getFileUploader({ acl: ACL.publicRead }, (_, file) =>
-      new S3Asset().getFileKey(file.fieldname)
-    )
+    const uploader = getFileUploader({ acl: ACL.publicRead }, (_, file) => {
+      console.log(file)
+      console.log(file.fieldname)
```
nicosantangelo, line 78: Logs!

src/Item/Item.router.ts
```diff
@@ -108,13 +108,14 @@ export class ItemRouter extends Router {
...
     try {
-      const uploader = getFileUploader({ acl: ACL.publicRead }, (_, file) =>
-        new S3Item(id).getFileKey(file.fieldname)
-      )
-      return utils.promisify<boolean>(uploader.any())
+      const uploader = getFileUploader({ acl: ACL.publicRead }, (_, file) => {
+        return new S3Item(id).getFileKey(file.fieldname)
+      })
+      const handler = utils.promisify<boolean>(uploader.any())
+      await handler(req, res)
```
nicosantangelo, line 118: up to you, but we could copy the way we do this on Asset.router by caching this handler before using it on `mount()`
It's effectively the same, but just to keep the same implementation

nicosantangelo (approved):
LGTM

PR #119, opened 2020-09-03
refactor: assets -> contents

cazala:


nicosantangelo (approved):

nicosantangelo:
@cazala We could update the README right? I can merge this PR and we can just update it in master later. However you think best

PR #125, opened 2020-09-24
feat: consolidate blockchain data

nicosantangelo:


src/Collection/Collection.router.ts
```diff
@@ -70,51 +96,58 @@ export class CollectionRouter extends Router {
...
-      ...collectionJSON,
-      eth_address
-    } as CollectionAttributes
-
-    if (id !== attributes.id) {
-      throw new HTTPError('The body and URL collection ids do not match', {
-        urlId: id,
-        bodyId: attributes.id
-      })
+      await new Collection(attributes).upsert()
```
cazala, line 139: shouldn't we `return` here instead of `await`? I don't think we do anything with the response on the frontend anyway, but this would change the content of the response compared to the what we have in master

cazala (approved):

src/Collection/Collection.router.ts
nicosantangelo: You're right!

PR #131, opened 2020-11-03
feat: land analytics

cazala:
`GET /analytics/weekly`


src/Analytics/Analytics.model.ts
```diff
@@ -0,0 +1,82 @@
...
+          min_session_time,
+          average_session_time,
+          max_session_time,
+          direct_users,
+          direct_sessions,
+          max_concurrent_users,
+          max_concurrent_users_time
+        } = result.rows[0]
+
+        weekly = {
```
nachomazzara, line 51: Small comment: Maybe we can return just this and remove the `weekly` variable.

src/Analytics/Analytics.model.ts
```diff
@@ -0,0 +1,82 @@
...
+          week,
+          title,
+          base,
+          users: parseInt(users, 10),
```
nachomazzara, line 55: Are we parsing it for some special reason?

nachomazzara (commented):
Awesome! thanks @cazala

src/Analytics/Analytics.model.ts
cazala: i didnt return it as is because the numeric values are strings

src/Analytics/Analytics.model.ts
cazala: becuase they are returned by the query as strings instead of numbers

nachomazzara: Yeah! I'm wondering if we can display as they are in the front-end as we don't need to manipulate them

nachomazzara (approved):

PR #135, opened 2020-11-18
refactor: remove nested collections and items

### LINKED ISSUE #132: Items within collections have isPublished=false even if collection's isPublished=true

cazala:
Fixes #132 

This PR removes the nested items within collections and the nested collections within items. This was causing issues when consolidating data between our database and the graph, also we were sending an unnecessary amount of data on each response from the server. The types have been also simplified a lot since I got rid of `CollectionWithItems`, `CollectionItem` and `ItemWithCollection`, and all of the `Partial<CollectionAttributes>` and `Partial<ItemAttributes>`, some of which were causing runtime errors due to the weak typings.

Now there are only two types for collections and two types for items: `ItemAttributes` and `CollectionAttributes` for stuff on our DB, and also that's the same type the server returns from its endpoints, and `ItemFragment` and `CollectionFragment` which are the types returned from the graph.


src/Item/Item.router.ts
```diff
@@ -41,14 +43,23 @@ export class ItemRouter extends Router {
...
+     * Returns an item
      */
     this.router.get(
       '/items/:id',
       withItemExists,
       server.handleRequest(this.getItem)
     )
 
+    /**
+     * Returns the item's of a collection
```
nicosantangelo, line 55:
```suggestion
     * Returns the items of a collection
```
right?

src/Collection/Collection.router.ts
```diff
@@ -71,46 +67,30 @@ export class CollectionRouter extends Router {
...
-        collection = {
-          ...collection,
-          ...remoteData.collection
-        }
-      }
+      return dbCollection
     }
 
-    return collection
+    return null
```
nicosantangelo, line 93: I think we can return undefined here? to be consistent with other endpoints (like the item endpoint below, which just returns `dbItem: Item | undefined`)

src/ethereum/api/Bridge.ts
```diff
@@ -1,65 +1,100 @@
...
+    const collectionIds = dbItems.reduce<string[]>((ids, dbItem) => {
+      if (dbItem.collection_id) {
+        ids.push(dbItem.collection_id)
+      }
+      return ids
+    }, [])
+
+    const dbResults = await Promise.all(
+      collectionIds.map(collectionId =>
+        Collection.findOne<CollectionAttributes>(collectionId)
```
nicosantangelo, line 44: we *could* improve performance here by adding a `findByIds` method, reducing this to a single query

src/ethereum/api/Bridge.ts
```diff
@@ -1,65 +1,100 @@
...
+    )
+    const dbCollections = dbResults.reduce((obj, result) => {
+      if (result) {
+        obj[result.id] = result
+      }
+      return obj
+    }, {} as Record<string, CollectionAttributes>)
+
+    // Get remote collections
+    const contractAddresses = Object.values(dbCollections).map(
```
nicosantangelo, line 55: You could just use dbResults here right?
```suggestion
    const contractAddresses = dbResults.map(
```

src/ethereum/api/Bridge.ts
```diff
@@ -1,65 +1,100 @@
...
+      dbCollection => dbCollection.contract_address
+    )
+    const remoteResults = await Promise.all(
+      contractAddresses.map(contractAddress =>
+        collectionAPI.fetchCollection(contractAddress)
```
nicosantangelo, line 60: we _could_ improve performance here by adding a `fetchByAddresses` method, reducing this to a single query. We might run into thegraph's limitations but I don't think we'll query >1k collections

src/ethereum/api/Bridge.ts
```diff
@@ -1,65 +1,100 @@
...
+      )
+    )
+    const remoteCollections = remoteResults.reduce((obj, result) => {
+      if (result) {
+        obj[result.id] = result
+      }
+      return obj
+    }, {} as Record<string, CollectionFragment>)
```
nicosantangelo, line 68: Maybe we can extract this to an instance method? It might not be worth it, it's just a suggestion but it could be something like

```ts
private function indexById<T extends { id: string }>(objects: T[]): Record<string, T> {
  return objects.reduce((obj, result) => {
    if (result) {
      obj[result.id] = result
    }
    return obj
  }, {} as Record<string, T>)
}

```

src/server.ts
```diff
@@ -73,6 +73,7 @@ async function startServer() {
     ) {
       // ignore timeouts and terminated connections
     } else {
+      console.error(err.message)
```
nicosantangelo, line 76: was this intentional? just in case :P

nicosantangelo (commented):
Lookin fresh

src/ethereum/api/Bridge.ts
cazala: `dbResults` is `(Collection | undefined)[]` that's why I didn't used that, the `undefined` values are filtered by the `if (result)` when I convert the array into object

src/server.ts
cazala: yes, i noticed that when uncaught errors occurred they didn't show up on the terminal without that (in my case I was trying to start the server in one terminal and there as another one already running on that same port, and it would die without logging anything)

src/ethereum/api/Bridge.ts
cazala: done 👍

src/ethereum/api/Bridge.ts
cazala: done 👍

src/ethereum/api/Bridge.ts
cazala: done 👍

src/Collection/Collection.router.ts
cazala: done 👍

src/Collection/Collection.model.ts
```diff
@@ -1,21 +1,14 @@
...
-  }
-
-  static findByEthAddressWithItems(ethAddress: string) {
-    return this.query<CollectionWithItems>(SQL`
-      ${CollectionQueries.selectWithItems()}
-        WHERE eth_address = ${ethAddress}`)
+  static findByEthAddress(ethAddress: string) {
+    return this.query<CollectionAttributes>(SQL`
+    SELECT *
+      FROM ${raw(Collection.tableName)}
```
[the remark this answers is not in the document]
cazala: done 👍

nicosantangelo (approved):

PR #136, opened 2020-11-20
feat: asset endpoints

### LINKED ISSUE #134: Add endpoint to return assets by id
`GET /assets/:id`

If possible would be great to have also a multi get endpoint:

```
GET /assets?id=1&id=2&id=3
```

cazala:
Closes #134 

New endpoint to return single asset

```
GET /assets/:id
```

New endpoint to return multiple assets

```
GET /assets?id=1&id=2&id=3
```

### DIFF
--- a/src/Asset/Asset.router.ts
+++ b/src/Asset/Asset.router.ts
@@ -103,16 +103,8 @@ export class AssetRouter extends Router {
   }
 
   private getAssets(req: Request) {
-    const ids = req.query.id
-      ? Array.isArray(req.query.id)
-        ? (req.query.id as string[])
-        : [req.query.id as string]
-      : []
-
-    if (ids.length === 0) {
-      throw new Error('You must provide at least one "id" via query params')
-    }
-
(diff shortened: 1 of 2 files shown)


src/Asset/Asset.router.ts
```diff
@@ -81,4 +96,23 @@ export class AssetRouter extends Router {
...
+    const id = server.extractFromReq(req, 'id')
+    return Asset.findOne(id)
+  }
+
+  private getAssets(req: Request) {
+    const ids = req.query.id
+      ? Array.isArray(req.query.id)
+        ? (req.query.id as string[])
+        : [req.query.id as string]
+      : []
```
nicosantangelo, line 110: What do you think about using `extractFromReq` here too?
something like
```ts
    const reqIds = server.extractFromReq<string | string[]>(req, 'ids')
    const ids: string[] = Array.isArray(reqIds) ? reqIds : [reqIds]
```

nicosantangelo (commented):
mini-comment

nicosantangelo (approved):

PR #170, opened 2021-02-26
Feat: allow managers and minters to operate

nachomazzara:
This PR allows managers and minters to fetch items/collections

### DIFF
--- a/src/ethereum/api/Bridge.ts
+++ b/src/ethereum/api/Bridge.ts
@@ -55,39 +55,31 @@ export class Bridge {
   ) {
     const items: ItemAttributes[] = []
 
-    // Get db collections from DB items
-    const collectionIds = dbItems.reduce<string[]>((ids, dbItem) => {
-      if (dbItem.collection_id) {
-        ids.push(dbItem.collection_id)
-      }
-      return ids
-    }, [])
-    const dbResults = await Collection.findByIds(collectionIds)
-    let dbCollections = this.indexById(dbResults)
+    // To avoid multiple queries to the db, we will fetch all the items that match the blockchain_id and their collections
+    // to filter them later
+    let remoteDBItems = await Item.findByBlockchainIdsAndContractAddresses(
+      remoteItems.map((remoteItem) => ({
+        blockchainId: remoteItem.blockchainId,
+        collectionAddress: remoteItem.collection.id,
+      }))
+    )
 
-    // Get db collections from remote items
-    const addresses = remoteItems
-      .filter(
-        (remoteItem) =>
-          !dbResults.some(
-            (collection) =>
-              collection.contract_address === remoteItem.collection.id
-          )
-      )
-      .map((item) => item.collection.id)
-    const collectionsDBFromremoteResults = await Collection.findByAddresses(
-      addresses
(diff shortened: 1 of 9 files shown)


src/ethereum/api/Bridge.ts
```diff
@@ -22,6 +24,28 @@ export class Bridge {
...
+      remoteCollectionCreators
+    )
+    for (const remoteCollection of remoteCollections) {
+      // Get only db collections that has not been added in the collection array
+      const dbCollection = dbCollectionsByRemotes.find(
+        (collection) =>
+          collection.contract_address === remoteCollection.id &&
+          !collections.some(
+            (collection) => collection.contract_address === remoteCollection.id
+          )
```
nicosantangelo, line 40: This seems...wrong. I don't really get what this condition is doing, but the first thing that seems odd is that you're doing a some (iterating) for each `dbCollectionsByRemotes`, but the result is always the same, as the collection variable is shadowing the one on the iteration (some)

At the very least, maybe we should move  this outside the find?

src/ethereum/api/Bridge.ts
```diff
@@ -31,51 +55,49 @@ export class Bridge {
...
-      (dbCollection) => dbCollection.contract_address
+    // Get db collections from DB items
+    const collectionIds = [...dbItems, ...remoteDBItems].reduce<string[]>(
+      (ids, dbItem) => {
+        if (dbItem.collection_id) {
+          ids.push(dbItem.collection_id)
+        }
+        return ids
+      },
+      []
```
nicosantangelo, line 75: I know it looks like I hate `.reduce` but...I do (?)

In all honesty, I think the method is not meant for these kinds of things and it just makes code way harder to read for no reason. For example if you want to go the functional route, you can:

```ts
[...dbItems, ...remoteDBItems].filter(item => !!item.collection_id).map(item => item.collection_id)
```

or just

```ts
const collectionIds = []
for(const item of [...dbItems, ...remoteDBItems]) {
  if(item.collection_id) {
    collectionIds.push(item.collection_id)
  }
}
```

In both these and another review where I suggested the same, you can see that doing a for...of usually is less code and way more readable

⚠️ This is my take on this, I have ways to justify it, but I'm open to chat about it

nicosantangelo (commented):
Looks good, a few comments

src/ethereum/api/Bridge.ts
```diff
@@ -22,6 +24,28 @@ export class Bridge {
       collections.push(collection)
     }
 
+    const remoteCollectionCreators = remoteCollections.map((collection) =>
+      getAddress(collection.creator)
```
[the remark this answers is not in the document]
nachomazzara: `getAddress` is used to validate an address and also to checksum it. Here, I checksum it because creators are being stored checksummed. What is really weird is why I am looking for creators instead of contract addresses.

I changed it to contract address.

src/ethereum/api/Bridge.ts
nachomazzara: Was not in my best dev momento. What a shame

nicosantangelo (approved):
LGTM, good job!

PR #171, opened 2021-02-26
feat: prevent items or collection to be deleted

nachomazzara:
Prevent collections to be updated or deleted if they are published
Prevent items to be deleted if they are published

### DIFF
--- a/src/Collection/Collection.router.ts
+++ b/src/Collection/Collection.router.ts
@@ -123,7 +123,7 @@ export class CollectionRouter extends Router {
 
       if (await this.isCollectionPublished(id)) {
         throw new HTTPError(
-          "The collection can't be updated because it is published",
+          "The collection is published. It can't be updated",
           {
             id,
           },
@@ -161,7 +161,7 @@ export class CollectionRouter extends Router {
 
     if (await this.isCollectionPublished(id)) {
       throw new HTTPError(
-        "The collection was published. It can't be deleted",
+        "The collection is published. It can't be deleted",
         {
           id,
         },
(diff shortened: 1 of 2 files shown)


src/Collection/Collection.router.ts
```diff
@@ -140,10 +150,37 @@ export class CollectionRouter extends Router {
 
   async deleteCollection(req: AuthRequest) {
     const id = server.extractFromReq(req, 'id')
+
+    if (await this.isCollectionPublished(id)) {
+      throw new HTTPError(
+        "The collection is published. It can't be deleted",
+        {
+          id,
+        },
```
nicosantangelo, line 159: can we indent both these objects as `{ id }` ?

src/Collection/Collection.router.ts
```diff
@@ -140,10 +150,37 @@ export class CollectionRouter extends Router {
...
+
     await Promise.all([
       Collection.delete({ id }),
       Item.delete({ collection_id: id }),
     ])
     return true
   }
+
+  async isCollectionPublished(collectionId: string) {
+    const dbCollection = await Collection.findOne<CollectionAttributes>({
```
nicosantangelo, line 172: I think you can just provider the collectionId to findOne right?

src/Item/Item.router.ts
```diff
@@ -187,6 +187,30 @@ export class ItemRouter extends Router {
 
   async deleteItem(req: AuthRequest) {
     const id = server.extractFromReq(req, 'id')
+
+    const dbItem = await Item.findOne<ItemAttributes>({ id })
+    const dbCollection = await Collection.findOne<CollectionAttributes>({
+      id: dbItem!.collection_id!,
```
nicosantangelo, line 193: I think you can just provide the id to findOne right?

nicosantangelo:
Added two more comments, post review.
Sorry!

nicosantangelo (approved):

PR #172, opened 2021-02-27
feat: prevent repeated collection names

nachomazzara:
Close decentraland/builder#1094

### DIFF
--- a/src/Collection/Collection.router.ts
+++ b/src/Collection/Collection.router.ts
@@ -49,7 +49,7 @@ export class CollectionRouter extends Router {
     this.router.put(
       '/collections/:id',
       withAuthentication,
-      server.handleRequest(this.upsertCollection.bind(this))
+      server.handleRequest(this.upsertCollection)
     )
 
     /**
@@ -113,7 +113,7 @@ export class CollectionRouter extends Router {
         )
       }
 
-      if (!(await this.isCollectionNameAvailable(collectionJSON))) {
+      if (await Collection.nameExist(collectionJSON.name.trim())) {
         throw new HTTPError(
           'Name already in use',
           { id, name: collectionJSON.name },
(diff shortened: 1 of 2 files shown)


src/Collection/Collection.router.ts
```diff
@@ -49,7 +49,7 @@ export class CollectionRouter extends Router {
     this.router.put(
       '/collections/:id',
       withAuthentication,
-      server.handleRequest(this.upsertCollection)
+      server.handleRequest(this.upsertCollection.bind(this))
```
nicosantangelo, line 52: The bind is not necessary, we're using the already bound method syntax `= () =>`

src/Collection/Collection.router.ts
```diff
@@ -146,4 +154,10 @@ export class CollectionRouter extends Router {
     ])
     return true
   }
+
+  async isCollectionNameAvailable(collection: CollectionAttributes) {
```
nicosantangelo, line 158: I think receiving the entre collection as an argument is a bit overkill. What do you think about just getting the name here?

If we had another piece of code trying to check for a name it'd have to build an entire collection out of it

src/Collection/Collection.router.ts
```diff
@@ -146,4 +154,10 @@ export class CollectionRouter extends Router {
...
+    const dbCollection = await Collection.findByName(collection.name.trim())
+
+    return dbCollection.length === 0
```
nicosantangelo, line 161: If this is the only use we're going to have for `findByName`, you can always speed things up by adding a `Collection.nameExists` which uses a count

nicosantangelo (commented):
Comments

src/Collection/Collection.router.ts
nachomazzara: Hmmm, we are not using it that way. 

I had to do it because I was getting:
Screen Shot 2021-03-01 at 21 37 17

src/Collection/Collection.router.ts
nachomazzara: Sounds reasonable

nicosantangelo (approved):
LGTM

PR #182, opened 2021-03-16
feat: support curation tab (committee)

nicosantangelo:


src/Collection/Collection.router.ts
```diff
@@ -64,10 +70,33 @@ export class CollectionRouter extends Router {
     )
   }
 
-  async getCollections(req: AuthRequest) {
-    const eth_address = req.auth.ethAddress
+  async getCollections(req: Request) {
+    let isPublished: boolean | undefined
+    try {
+      isPublished = new RequestParameters(req).getBoolean('isPublished')
```
nachomazzara, line 76: As we called `eth_address` the param for `getAddressCollection`. Should we need to call this `is_published`?

nicosantangelo: You are be right, this is annoyingly inconsistent. I think it makes sense to call the other variable ethAddress to keep the cameCase nomenclature. I'll fix it on both routers 👍

nicosantangelo: well darn, this:
`const eth_address = req.auth.ethAddress`
is scattered in the entire codebase. I'll change it to is_published like you said

PR #185, opened 2021-03-30
feat: add forum link to collections

nicosantangelo:


src/Forum/Forum.router.ts
```diff
@@ -0,0 +1,57 @@
+// import { Request, Response } from 'express'
```
nachomazzara, line 1: Oops!

src/Forum/client.ts
```diff
@@ -0,0 +1,49 @@
...
+}
+
+export type CreateError = {
+  action: string
+  errors: string[]
+}
+
+type CreateResponse = CreateSuccess | CreateError
+
+export const FORUM_URL = env.get('REACT_APP_FORUM_URL', '')
```
nachomazzara, line 23: Is this variable being used outside this file?

src/Forum/client.ts
```diff
@@ -0,0 +1,49 @@
+import fetch, { Response } from 'node-fetch'
+import { env } from 'decentraland-commons'
+import { ForumPost } from './Forum.types'
+
+export type CreateSuccess = {
```
nachomazzara, line 5: Should we need to keep these types here or on `Forum.types.ts`?

src/Forum/client.ts
nicosantangelo: Not really! I'll delete the export

src/Forum/client.ts
nicosantangelo: Good catch, probably on `.types.ts`, I meant to have them there and forgot

PR #189, opened 2021-04-05
feat: fetch rarities from graph

nachomazzara:


src/Committee/Committee.router.ts
```diff
@@ -7,7 +7,7 @@ import { collectionAPI } from '../ethereum/api/collection'
 export class CommitteeRouter extends Router {
   mount() {
     /**
-     * Returns the addresses fo the current committee
+     * Returns the addresses for the current committee
```
nicosantangelo, line 10: image

nicosantangelo (approved):
Awesome !

PR #197, opened 2021-04-12
chore: builder in world added field

Suduck:
### What
This change adds a new field to that links the builder project to land.

This is needed since we need to know from builder in world if there is any project associated with the current land. 
If there is we will load it. 

In case we are not able to find any project associated with the land, we will create one linked to the coordinates.

The builder will continue to work with a null in that field


src/Manifest/Manifest.router.ts
```diff
@@ -83,6 +92,26 @@ export class ManifestRouter extends Router {
     }
   }
 
+  async getProjectManifestByCoordinates(req: AuthRequest) {
+    const created_location = server.extractFromReq(req, 'land')
+    const eth_address = req.auth.ethAddress
+
+    const project : any = await Project.findOne({
```
nicosantangelo, line 99: This works! but we can type it better. There are two options:
- You could replace `any` here with the actual type, wich would be `ProjectAttributes | undefined` (from `/Project`)
- Or, better still, use the generic on `findOne` like so: `Project.findOne<ProjectAttributes>(...)`

src/Manifest/Manifest.router.ts
```diff
@@ -83,6 +92,26 @@ export class ManifestRouter extends Router {
...
+    })
+    if (project) {
+      const body = await new S3Project(project.id).readFileBody(
+        MANIFEST_FILENAME
+      )
+      if (body) {
+        return JSON.parse(body.toString())
+      }
+    }
+    return 'false'
```
nicosantangelo, line 112: We could just return a boolean here, right? it'd be correctly handled as we're ultimately returning a JSON.

src/Manifest/Manifest.router.ts
```diff
@@ -83,6 +92,26 @@ export class ManifestRouter extends Router {
     }
   }
 
+  async getProjectManifestByCoordinates(req: AuthRequest) {
+    const created_location = server.extractFromReq(req, 'land')
```
nicosantangelo, line 96: I'm seeing three different concepts to describe what it seems to be the same thing:
- Coordinates
- land
- created_location

Maybe we could just settle on a name and use that everywhere? coordinates seems like a good name, or maybe just `landId`?

src/Manifest/Manifest.router.ts
```diff
@@ -36,6 +36,15 @@ export class ManifestRouter extends Router {
       server.handleRequest(this.getProjectManifest)
     )
 
+    /**
+     * Returns the manifest of a land associated project  (Builder In World)
+     */
+    this.router.get(
+      '/projects/:land/manifestFromCoordinates',
```
nicosantangelo, line 43: Careful here! remember that using REST means doing_something like_:
`resource/ID/action`
but :land here is not indexing project.

Maybe we can add a query string filter to the already existing `projects/:id/manifest` route? @cazala

README.md
```diff
@@ -26,14 +30,14 @@ npm start
 
 # Rewrites to S3
 
-- `GET /storage/assetPacks/:filename` => https://s3.amazonaws.com/nico.decentraland.zone/asset_packs/:filename (for the thumbnail, :id.png)
-- `GET /storage/contents/:hash` => https://s3.amazonaws.com/nico.decentraland.zone/contents/:hash
+- `GET /v1/storage/assetPacks/:filename` => https://s3.amazonaws.com/nico.decentraland.zone/asset_packs/:filename (for the thumbnail, :id.png)
```
nicosantangelo, line 33: I think that if we're going to add `v1` here, we should add a little disclaimer up top explaining that the value comes from the .env's `API_VERSION` variable. In case we change it later. What do you think?

nicosantangelo (changes requested):
Great job 🔥 !

Don't forget linting the code too! Found a few errors running `npm run lint`:
image

We usually just let our editor of choice run prettier on save to avoid this

src/Manifest/Manifest.router.ts
Suduck: Changed for created_location, I think since the database field have this name it will be more correct

src/Manifest/Manifest.router.ts
Suduck: Sure let me know how do you want it and I will change it! An example to help will be cool!

cazala: The thing with `/projects/:id/mainfest` is that it requires you to know the project ID, which is not the case here. In order to keep this RESTful, maybe we can just use the endpoint `/manifests` with a `created_location` query param, like `/manifests?created_location=10,10`.

nicosantangelo: oh yes, thought of that afterwards but wasn't sure if the id was available. I like the idea of having a `/manifests` endpoint with a query param.
The only problem I see with it is that you **have** to have access to the project(s) returned by it. So the idea would be to return all manifests you have access to, filtering by the query string.

Regarding [this comment](https://github.com/decentraland/builder-server/pull/197#discussion_r611717182) I see that it's [still a thing](https://github.com/decentraland/builder-server/pull/197/files#diff-536f13249592afdba56fe4ccd5ba79296b564dcc392bbd5049f46509ece9e6e0R95) (with the method name). What do you guys think of using `coordinates` or `land_coordinates` everywhere?
I think coordinates is already a concept we use in parcel-related projects

@Suduck @cazala

Suduck: It is fine by me! let me know which one you prefer and I will change it in all parts!

nicosantangelo: Ok after talking a bit with juan, it's a hard name to do, but what do you think of `creation_coords` ? and to use that for the query string @Suduck 

if this is time-pressing we can go with the current implementation and change it later

Suduck: Sounds good to me! I will change it now since it will be easier because if we change it later we will need to sync kernel and unity master branches at the same time to ensure it is working. 

I will make the changes and ask for a revision again!

nicosantangelo: Thanks a lot! sorry for the back and forth!

src/Manifest/Manifest.router.ts
```diff
@@ -77,12 +87,34 @@ export class ManifestRouter extends Router {
...
+        const body = await new S3Project(project.id).readFileBody(
+          MANIFEST_FILENAME
+        )
+
+        if (body) manifests.push(JSON.parse(body.toString()))
+      }
+
+      if (manifests) return manifests
+    }
+    return false
```
cazala, line 115: maybe we could return `[]` instead of `false` here? We could move the `const manifest: Manifest[] = []` outside, before the `if (projects) { ... }` block, and just `return manifests` at the end of the function.

Also, this is missing the `?creation_coords` query param, right? Should we add it before returning the `manifests` array, something like:

```
const creationCoords: string | null = server.extractFromReq(req, 'creation_coords')
if (creationCoords) {
  const project = projects.items.find(project => project.creation_coords === creationCoords)
  const manifest = manifests.find(manifest => manifest.project.id === project.id) || null
  return manifest
}
```

Suduck: Ok, I will change it to return an empty array.

For the creation coords part, it is not needed,

with this part of the code
`
    const projectSearcher = new SearchableProject(req)
    const projects = await projectSearcher.searchByEthAddress(eth_address)`
    
    You don't need to specify for creation coords, you just can filter by any parameter that you want from the project

cazala: ahh okay, i didn't realize that last part, cool :)

cazala (approved):

src/Manifest/Manifest.router.ts
```diff
@@ -77,12 +87,34 @@ export class ManifestRouter extends Router {
...
       return JSON.parse(body.toString())
     }
   }
 
+  async getManifests(req: AuthRequest) {
+    let eth_address = req.auth.ethAddress
+
+    const projectSearcher = new SearchableProject(req)
+    const projects = await projectSearcher.searchByEthAddress(eth_address)
+    let manifests: any[] = []
```
nicosantangelo, line 102: I think juanca touched on this on his comment but I'd change this from any to the actual type, is it ManifestAttributes? (from Manifest.types.ts)

src/Project/SearchableProject.ts
```diff
@@ -0,0 +1,31 @@
...
+    const conditions = new SearchableConditions<ProjectAttributes>(
+      requestParameters,
+      { eq: searchableProjectProperties }
+    )
+    conditions.addExtras('eq', { eth_address })
+    conditions.addExtras('eq', { is_deleted: false })
+
+    return searchableProject.search(parameters, conditions)
+  }
+}
```
nicosantangelo, line 31: awesome ❤️ 
we can change it in `Pool.router#getPools` too! (from the comment [here](https://github.com/decentraland/builder-server/pull/197/files#diff-98e02be08a7016d8e5703d4e71cbd20510c54f36706cc819866569aefc8685e1L118)

src/Manifest/Manifest.router.ts
```diff
@@ -36,6 +37,15 @@ export class ManifestRouter extends Router {
       server.handleRequest(this.getProjectManifest)
     )
 
+    /**
+     * Returns the manifest of a land coordinates associated project  (Builder In World)
+     */
+    this.router.get(
+      '/manifests',
+      withAuthentication,
```
nicosantangelo, line 45: **Question** with this impelementation, an authenticated user can ask for other users's manifests. Is this correct? the endpoint below, has [withProjectAuthorization](https://github.com/decentraland/builder-server/pull/197/files#diff-536f13249592afdba56fe4ccd5ba79296b564dcc392bbd5049f46509ece9e6e0L73) which checks that the authorised user owns the project

nicosantangelo (commented):
Awesomee

src/Manifest/Manifest.router.ts
Suduck: But it will work if we don't have the ID? I am looking at the code but it checks for ID, in my case, there won't be an ID, since I am looking for the new field.

Should I add to the `withProjectAuthorization `to check for any properties that are in the request?

nicosantangelo (approved):

PR #201, opened 2021-04-19
feat: add basic metrics

nachomazzara:


src/common/ExpressApp.ts
```diff
@@ -44,6 +47,48 @@ export class ExpressApp {
     return this
   }
 
+  useMetrics() {
+    const base = createTestMetricsComponent(getDefaultHttpMetrics())
+    const register = base.register
+
+    this.router.get('/metrics', async (_: any, res: express.Response) => {
```
nicosantangelo, line 54: _ could be express.Request even if you're not using it, to avoid an `any`

nicosantangelo (approved):
LGTM

PR #231, opened 2021-06-29
feat: remove projects by coords

Suduck:
Builder in world needs the ability to unlink the builder projects to the land coords when it is published.

with this PR we create an endpoint to remove the link of all projects when it is published so no more projects are associated to the land but they are still accessible in builder

### DIFF
--- a/src/Project/Project.router.ts
+++ b/src/Project/Project.router.ts
@@ -83,7 +83,7 @@ export class ProjectRouter extends Router {
      * Update all projects with coords to null
      */
     this.router.delete(
-      '/projects/coords/:coords',
+      '/projects/:coords/coords',
       withAuthentication,
       server.handleRequest(this.removeCoordsFromProjects)
     )
--- a/src/Project/Project.router.ts
+++ b/src/Project/Project.router.ts
@@ -79,6 +79,15 @@ export class ProjectRouter extends Router {
       server.handleRequest(this.deleteProject)
     )
 
+    /**
+     * Update all projects with coords to null
+     */
+    this.router.post(
+      '/projects/coords/:coords',
+      withAuthentication,
+      server.handleRequest(this.removeCoordsFromProjects)
+    )
+
     this.router.get(
       '/projects/:id/public',
       withProjectExistsAndIsPublic,
@@ -156,6 +165,13 @@ export class ProjectRouter extends Router {
   async deleteProject(req: AuthRequest) {
     const id = server.extractFromReq(req, 'id')
     await Project.update({ updated_at: new Date(), is_deleted: true }, { id })
+    return true
+  }
+
+  async removeCoordsFromProjects(req: AuthRequest) {
+    const eth_address = req.auth.ethAddress
+    const creation_coords = server.extractFromReq(req, 'coords')
+    await Project.update({ updated_at: new Date(), creation_coords: undefined }, { creation_coords, eth_address })
     return true
   }
 
--- a/src/Project/Project.router.ts
+++ b/src/Project/Project.router.ts
@@ -82,7 +82,7 @@ export class ProjectRouter extends Router {
     /**
      * Update all projects with coords to null
      */
-    this.router.post(
+    this.router.delete(
       '/projects/coords/:coords',
       withAuthentication,
       server.handleRequest(this.removeCoordsFromProjects)
--- a/src/Project/Project.router.ts
+++ b/src/Project/Project.router.ts
@@ -171,7 +171,10 @@ export class ProjectRouter extends Router {
   async removeCoordsFromProjects(req: AuthRequest) {
     const eth_address = req.auth.ethAddress
     const creation_coords = server.extractFromReq(req, 'coords')
-    await Project.update({ updated_at: new Date(), creation_coords: undefined }, { creation_coords, eth_address })
+    await Project.update(
+      { updated_at: new Date(), creation_coords: undefined },
+      { creation_coords, eth_address }
+    )
     return true
   }
 
--- a/src/middleware/authentication.ts
+++ b/src/middleware/authentication.ts
@@ -3,7 +3,7 @@ import { AuthLink } from 'dcl-crypto'
 import { server } from 'decentraland-server'
 import { STATUS_CODES } from '../common/HTTPError'
 import { AuthRequestLegacy } from './authentication-legacy'
-//import { peerAPI } from '../ethereum/api/peer'
+import { peerAPI } from '../ethereum/api/peer'
 
 const AUTH_CHAIN_HEADER_PREFIX = 'x-identity-auth-chain-'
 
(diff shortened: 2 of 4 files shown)


src/Project/Project.router.ts
```diff
@@ -79,6 +79,15 @@ export class ProjectRouter extends Router {
       server.handleRequest(this.deleteProject)
     )
 
+    /**
+     * Update all projects with coords to null
+     */
+    this.router.post(
```
menduz, line 85:
```suggestion
    this.router.delete(
```

May this be a `delete` instead? Post implies creation of resources

menduz (changes requested):

src/Project/Project.router.ts
Suduck: Delete usually implies that the row is deleted of marked as deleted, in this case the only thing we do is unlink the project to a land, but the project is still accessible by the builder and in a future, by the builder in world, so I see it as an "update". 

However, if you want I can set it to delete

menduz: If that's the case, I'd prefer to use the delete. It is more aligned with the concept of entities of the REST apis. you are deleting the `coord` property of the `project` entity

nicosantangelo: Would it make sense to just use the already existing `PUT /project/:id` ? I assume the `id` is not available right?

I think I prefer either PUT/DELETE (also because this method in particular is Idempotent)

Suduck: Yep, we don't have the `id ` of all of your projects. Changed to Delete!

src/Project/Project.router.ts
```diff
@@ -79,6 +79,15 @@ export class ProjectRouter extends Router {
       server.handleRequest(this.deleteProject)
     )
 
+    /**
+     * Update all projects with coords to null
+     */
+    this.router.delete(
+      '/projects/coords/:coords',
+      withAuthentication,
```
menduz, line 87: aren't the following middlewares missing?

```
      withProjectExists,
      withProjectAuthorization
```

menduz (changes requested):

src/Project/Project.router.ts
Suduck: As far as I know, those methods are used when you have the project ID to verify that the project exist. It also verify that you have the authorization to the project (by ID). This is done before doing the request, and since we don't have the project ID at that moment, we can't use these 

In this case, as we don't have the project ID, we update all the project associated with the eth_address that firm the request and removes the link to the coords. 

So it should be save because you won't be able to update others project unless you have their private key, and if you have it you could do all the calls anyways.

menduz: I will defer to @nicosantangelo on this one since he has full context about that, from what I see in the code it may be a security issue but I'm lacking detailed understanding of this code to not seeing it as a risk.

To be more specific, I don't see _which_ project are you updating when you execute:
```ts
    await Project.update(
      { updated_at: new Date(), creation_coords: undefined },
      { creation_coords, eth_address }
    )
```

Suduck: Sure! From my understanding, I am updating the projects that have a link with the coords that you send to the API and are created from the same eth_address that are requesting the call

src/Project/Project.router.ts
```diff
@@ -79,6 +79,15 @@ export class ProjectRouter extends Router {
       server.handleRequest(this.deleteProject)
     )
 
+    /**
+     * Update all projects with coords to null
+     */
+    this.router.delete(
+      '/projects/coords/:coords',
```
nicosantangelo, line 86: I think we could make this a bit more RESTful if we rename it to `/projects/:coords/coords` right? maybe we can think of an action as the last path instead of just coords, but at least the DELETE verb is semantic

nicosantangelo (commented):
Mini comment, sorry for the delay

src/Project/Project.router.ts
Suduck: Sure! It has been changed!

src/Project/Project.router.ts
```diff
@@ -83,7 +83,7 @@ export class ProjectRouter extends Router {
      * Update all projects with coords to null
      */
     this.router.delete(
-      '/projects/coords/:coords',
+      '/projects/:coords/coords',
```
menduz, line 86: Please use consistent URLs, all other endpoints use the `/projects/:id` prefix. This would make the API inconsistent and hard to both understand and maintain

menduz (changes requested):

src/Project/Project.router.ts
Suduck: I have made this changes because @nicosantangelo  asked me to do it. Before the last commit it was this way, please discuss about which one you prefer and I will implement it!

nicosantangelo: I share the sentiment, the problem here (I think) is that this entire interaction was not planned beforehand and the explorer does not have the id available to make requests. I'm all for using :id here but I don't know what that change'd imply

Suduck: I don't understand what you mean there. Are you suggesting creating a new router? Or I should make an API call to get all the projects with IDs and then do a call to removes the coords of all IDs? (That will be 2 API calls instead of 1, increasing the use of the server)

menduz: Ok, I asked for some context offline and I got that this is so because you don't have the project id. 

If that is the case, I recommend to seek for consistency and use a proper `/project_by_coord/:coords/coords` url and call it a day, to not mess up with the routes.

I also want to raise the concern for future developments: this kind of design considerations should arise in early stages of thinking, designing and reaching consensus of the features. Not in pull requests. Because this process is not beneficial for anybody, and the end solution ends up being far from ideal, breaking conventions and looking like a "patch to make it work".

Suduck: Done! Ok. Next time I will do a meeting to ensure that we are following the same path

src/middleware/authentication.ts
```diff
@@ -3,7 +3,7 @@ import { AuthLink } from 'dcl-crypto'
 import { server } from 'decentraland-server'
 import { STATUS_CODES } from '../common/HTTPError'
 import { AuthRequestLegacy } from './authentication-legacy'
-import { peerAPI } from '../ethereum/api/peer'
+//import { peerAPI } from '../ethereum/api/peer'
```
nachomazzara, line 6: Oops

src/middleware/authentication.ts
```diff
@@ -70,7 +70,7 @@ async function decodeAuthChain(req: Request): Promise<string> {
   } else {
     ethAddress = authChain[0].payload
 
-    if (!ethAddress) {
+   /* if (!ethAddress) {
```
nachomazzara, line 73: Oops

Suduck: Oops, change done!

src/middleware/authentication.ts
Suduck: My bad, fixed!

menduz (approved):

PR #232, opened 2021-06-29
feat: added fallback to check if collection is published

### LINKED ISSUE #229: Check if the collection is deployed by using `getCode`
We should do this if the subgraph returns that is not deployed.

cazala:
Closes #229


nachomazzara (approved):
LGTM

src/utils/eth.ts
```diff
@@ -0,0 +1,17 @@
...
+import { isContract } from 'decentraland-transactions'
+
+export const MATIC_CHAIN_ID = env.get(
+  'MATIC_CHAIN_ID',
+  ChainId.MATIC_MAINNET
+) as ChainId
+
+export async function isPublished(collectionAddress: string) {
+  const rpcUrl = RPC_URLS[MATIC_CHAIN_ID]
+  const provider = new providers.JsonRpcProvider(rpcUrl)
```
nicosantangelo, line 14: are there any benefits to having this as a local variable instead of re-instantiating on each request?

src/utils/eth.ts
```diff
@@ -0,0 +1,17 @@
...
+  const result = await isContract(provider, collectionAddress)
+  return result
```
nicosantangelo, line 16: We could just return isContract here if you want

src/Collection/Collection.router.ts
```diff
@@ -105,7 +105,7 @@ export class CollectionRouter extends Router {
       typeof is_published === 'undefined'
         ? Collection.find<CollectionAttributes>()
         : Collection.find<CollectionAttributes>({ is_published }),
-      collectionAPI.fetchCollections(),
```
nicosantangelo, line 108: I think there's been a mixup with prettier versions as, when I re-save this file I get the changes:
image

src/Collection/Collection.router.ts
```diff
@@ -265,6 +265,14 @@ export class CollectionRouter extends Router {
       dbCollection.contract_address
     )
 
-    return !!remoteCollection
+    // Fallback: check against the blockchain, in case the subgraph is lagging
+    if (!remoteCollection) {
+      const result = await isPublished(dbCollection.contract_address)
```
nicosantangelo, line 270: We could return !(await isPublished()) if you want to skip an if

nicosantangelo (commented):
Few silly comments, but we do need to check the linting problem

src/Collection/Collection.router.ts
cazala: mm could be related to this? https://github.com/decentraland/builder-server/blob/master/tslint.json#L4

src/utils/eth.ts
cazala: done 👍

src/utils/eth.ts
cazala: nope, i moved it outside 👍

src/Collection/Collection.router.ts
cazala: mm yes I could, but I think it's cleaner to let the final return to always be `true` and let the `if`s branch out if the return should be `false`.

src/Collection/Collection.router.ts
cazala: im getting these when i `npm ls prettier`:
```
builder-server@2.10.0 /Users/juancazala/code/builder-server
├─┬ dcl-catalyst-client@4.0.5
│ └─┬ dcl-catalyst-commons@3.0.1
│   └── prettier@2.2.1 deduped
└── prettier@2.2.1
```

PR #236, opened 2021-07-12
feat: update asset ids to be unique. update constraints

### LINKED ISSUE #234: Fix duplicated Asset ids on the database
Some assets in the builder have repeated ids (like `45299383e220b5f5675f55e46d6e9d1cdc57c35782d90a5a713bfed95e15207f`). This is because they're being imported via de `seed` script from an old batch of assets that didn't use a UUID and also because the table has a comound unique id with `(id, asset_pack_id)` to compensate for this fact

This is creating problems with other parts of Decentraland, so we need to update it. The idea then is to:
- Update ids so they have a uuid. This means updating the jsons used to populate the database in the `seed` script and also updating the database itself using a migration
- Update the database to have unique ids instead of a compound key using a migration

nicosantangelo:
## To test

```
npm run migrate up
npm run build
npm run seed
```

Closes #234


scripts/seed/1567613264447_data/83564d63-6e14-4469-abe6-60680391183c.json
```diff
@@ -1,12 +1,12 @@
 {
   "ok": true,
   "data": {
-    "id": "83564d63-6e14-4469-abe6-60680391183c",
+    "i": "83564d63-6e14-4469-abe6-60680391183c",
```
nachomazzara, line 4: iiiiiiiiiiiiiiiiii

migrations/1626119696313_update-asset-ids.ts
```diff
@@ -0,0 +1,30 @@
...
+	pgm.sql(`TRUNCATE TABLE ${tableName};`)
+
+	pgm.dropConstraint(tableName, 'assets_pkey')
+	pgm.addConstraint(tableName, 'assets_pkey', { primaryKey: 'id' })
+
+	// pgm won't infer this name correctly, so we need to nudge it a little
+	pgm.dropIndex(tableName, columns, {
+		name: 'assets_id_asset_pack_id_unique_index',
+	})
+	pgm.addIndex(tableName, 'id', { unique: true })
```
nachomazzara, line 17: Should we need update the `id` to be `uuid`?

package.json
```diff
@@ -12,7 +12,7 @@
     "watch:start": "nodemon src/server.ts",
     "migrate": "ts-node ./migrations/migrate.ts",
     "migrate:docker": "./migrations/node-pg-migrate --database-url-var CONNECTION_STRING --migration-file-language ts --migrations-dir /app/migrations --ignore-pattern '\\..*|.*migrate(.ts)?'",
-    "seed": "cp -r ./scripts/seed/*_data ./dist/scripts/seed && node ./dist/scripts/seed/index.js",
+    "seed": "npm run build && cp -r ./scripts/seed/*_data ./dist/scripts/seed && node ./dist/scripts/seed/index.js",
```
nachomazzara, line 15: I'm not sure about building the project here as we run the seed every time [we deploy a new version](https://github.com/decentraland/builder-server/blob/master/entrypoint.sh#L4). At that point, the project was already built.

Moreover, we should find a way to prevent migration and running the seed every time the project is deployed.

migrations/1626119696313_update-asset-ids.ts
nicosantangelo: Good catch! I think so, yes

package.json
nicosantangelo: Hm you're right. Thought about it when adding the build here, but the build was fast so didn't make much of it. It's weird to have the local project, run seed and it exploding, but I agree that this is not the solution

cazala (approved):

PR #240, opened 2021-07-20
fix: smart items uuids

nicosantangelo:


migrations/1626794554450_add-legacy-id-to-assets.ts
```diff
@@ -0,0 +1,16 @@
+import { MigrationBuilder } from 'node-pg-migrate'
+import { Asset } from '../src/Asset'
+
+const tableName = Asset.tableName
+
+export const up = (pgm: MigrationBuilder) => {
+  pgm.addColumn(tableName, {
+    legacy_id: { type: 'TEXT' },
```
cazala, line 8: Shouldn't this column be nullable? After the migration all the assets should have a legacy id, but the new ones created afterwards should not.

PR #241, opened 2021-07-23
feat: publish endpoint

nicosantangelo:
`POST  /collection/:id/publish`

### DIFF
--- a/src/Collection/Collection.router.ts
+++ b/src/Collection/Collection.router.ts
@@ -11,6 +11,7 @@ import {
 } from '../middleware'
 import { collectionAPI } from '../ethereum/api/collection'
 import { Bridge } from '../ethereum/api/Bridge'
+import { peerAPI, Wearable } from '../ethereum/api/peer'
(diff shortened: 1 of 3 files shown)


nachomazzara (approved):

src/Collection/Collection.router.ts
```diff
@@ -162,6 +174,69 @@ export class CollectionRouter extends Router {
...
+
+      await Promise.all(updates)
+    }
+
+    return {
+      collection: await Bridge.consolidateCollections(
+        [dbCollection!],
+        [remoteCollection]
+      ),
+      items: await Bridge.consolidateItems(dbItems, remoteItems, catalystItems),
```
nachomazzara, line 236: Should we need to use `items` here?

nicosantangelo: In this particular case it's the same result, but it's much better if we end up changing the endpoint. Updated!

nachomazzara (approved):

PR #244, opened 2021-08-04
feat: Log retries and output a conflict error when the link exists

LautaroPetaccio:
This PR logs the forum post retries and returns a 409 when there's a link for the forum already created for a collection.
It also fixes the entrypoint to avoid re-building the service when starting it.

### DIFF
--- a/entrypoint.sh
+++ b/entrypoint.sh
@@ -2,4 +2,4 @@
 
 npm run migrate:docker up || exit 1
 npm run seed || exit 1
-npm run start || exit 1
+node ./dist/src/server.js || exit 1
--- a/package.json
+++ b/package.json
@@ -8,7 +8,7 @@
     "lint:fix": "tslint -p tsconfig.json --fix",
     "build": "tsc -p tsconfig.json",
     "watch:build": "tsc -p tsconfig.json --watch",
-    "start": "npm run build && node ./dist/src/server.js",
+    "start": "ts-node src/server.ts",
     "watch:start": "nodemon src/server.ts",
     "migrate": "ts-node ./migrations/migrate.ts",
     "migrate:docker": "./migrations/node-pg-migrate --database-url-var CONNECTION_STRING --migration-file-language ts --migrations-dir /app/migrations --ignore-pattern '\\..*|.*migrate(.ts)?'",
@@ -24,6 +24,9 @@
     "semi": false,
     "singleQuote": true,
     "printWidth": 80
+  },
+  "engines": {
+    "node": ">=10 <=12"
   },
   "dependencies": {
     "@apollo/client": "^3.3.9",
--- a/src/Forum/Forum.router.ts
+++ b/src/Forum/Forum.router.ts
@@ -8,17 +8,18 @@ import { ExpressApp } from '../common/ExpressApp'
 import { withModelExists, withModelAuthorization } from '../middleware'
 import { withAuthentication, AuthRequest } from '../middleware/authentication'
 import { Collection, CollectionAttributes } from '../Collection'
+import { MetricDeclarations } from '../MetricsDeclarations'
 import { createPost } from './client'
 import { ForumPost, forumPostSchema } from './Forum.types'
 
 const validator = getValidator()
 
 export class ForumRouter extends Router {
-  readonly metrics: IMetricsComponent<'dcl_published_collection_forum_post_failed'>
+  readonly metrics: IMetricsComponent<MetricDeclarations>
 
   constructor(
     router: ExpressApp | express.Router,
-    metrics: IMetricsComponent<'dcl_published_collection_forum_post_failed'>
+    metrics: IMetricsComponent<MetricDeclarations>
   ) {
     super(router)
     this.metrics = metrics
@@ -35,7 +36,7 @@ export class ForumRouter extends Router {
       withAuthentication,
       withCollectionExists,
       withCollectionAuthorization,
-      server.handleRequest(this.post)
+      server.handleRequest(this.post.bind(this))
     )
   }
 
--- a/src/Forum/Forum.router.ts
+++ b/src/Forum/Forum.router.ts
@@ -1,8 +1,10 @@
 import { server } from 'decentraland-server'
-
+import express from 'express'
+import { IMetricsComponent } from '@well-known-components/interfaces'
 import { Router } from '../common/Router'
 import { HTTPError, STATUS_CODES } from '../common/HTTPError'
 import { getValidator } from '../utils/validator'
+import { ExpressApp } from '../common/ExpressApp'
 import { withModelExists, withModelAuthorization } from '../middleware'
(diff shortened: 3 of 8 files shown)


src/MetricsDeclarations.ts
```diff
@@ -0,0 +1,13 @@
+import { IMetricsComponent } from '@well-known-components/interfaces'
+import { validateMetricsDeclaration } from '@well-known-components/metrics'
+
+export const metricDeclarations = {
+  dcl_published_collection_forum_post_failed: {
+    help: 'Count failed published collection forum posts',
+    type: IMetricsComponent.CounterType,
+    labelNames: ['forum'],
```
nachomazzara, line 8: I think that we can remove the label here, right? We are not using it

src/server.ts
```diff
@@ -36,6 +38,8 @@ app
   .useCORS(CORS_ORIGIN, CORS_METHOD)
   .useMetrics()
 
+const metrics = createTestMetricsComponent(metricDeclarations)
```
nachomazzara, line 41: What do you think of moving [this](https://github.com/decentraland/builder-server/blob/master/src/common/ExpressApp.ts#L54) here and use `metrics` as a parameter of `useMetrics`

src/server.ts
```diff
@@ -49,7 +53,7 @@ new ItemRouter(app).mount()
 new CollectionRouter(app).mount()
 new CommitteeRouter(app).mount()
 new RarityRouter(app).mount()
-new ForumRouter(app).mount()
+new ForumRouter(app, metrics).mount()
```
nachomazzara, line 56: Not sure if it is possible but maybe we can avoid override the Router's constructor and send `metrics` at `mount`

nachomazzara (changes requested):
Awesome! thanks @LautaroPetaccio

src/server.ts
LautaroPetaccio: Makes sense!, we can merge both metrics

src/server.ts
LautaroPetaccio: The reason behind not doing that is that we want to set an instance of the metric component once that can be used by every route handler defined in the RouterClass.
There are three alternatives:
- We set it in the constructor once and use it in all the routers or methods in the class.
- We pass it to the mount method and propagate it to each of the routes being created by doing something like: `server.handleRequest((req, res) => this.post(req, res, metrics))`
- We pass it to the mount method and set it on each mount, making the metrics mutable, so any handler or method in the class can use it (I wouldn't to this, as metrics isn't meant to be mutable)

src/MetricsDeclarations.ts
LautaroPetaccio: Done!

src/server.ts
nachomazzara: Great! thanks for the explanation. Another alternative is to create and use a middleware to inject the metrics like `withMetrics` and use it like

```ts
this.router.post(
      '/collections/:id/post',
      withAuthentication,
      withCollectionExists,
      withCollectionAuthorization,
      withMetrics,
      server.handleRequest(this.post)
    )
```

I will leave this to you & @nicosantangelo.

nicosantangelo: Because we're borrowing the metrics from well-known-components, I think it makes sense to have this as an instance of our routers. This is because it's a complex object being created once on the `server.ts` and re-used later.

That said, what do you think of generalizing this to make our Routers metric aware? I'm thinking we can move the logic of accepting a `metricsComponent` to the abstract class and that we could type the strings that make our metrics like this:

```ts
export enum MetricKeys {
  FORUM_POST_FAILED = 'dcl_published_collection_forum_post_failed',
  FORUM_POST_ALREADY_EXISTS = 'dcl_published_collection_forum_post_already_exists',
}

export const metricDeclarations = {
  [MetricKeys.FORUM_POST_FAILED]: {
    (...)
```

and then either have:

```ts
export class Router<M extends string = string> {
  protected router: express.Router
  protected metrics?: IMetricsComponent<M>
  (...)

// or

export class MetricsRouter<M extends string = string> extends Router {
  protected metrics?: IMetricsComponent<M>
```

What do you think @LautaroPetaccio 

Lastly, but importantly, I think the increment on ForumRouter, that's using `'dcl_published_collection_forum_post_already_exists'` it's not correctly typed, as it's missing from the `IMetricsComponent` generic definition

LautaroPetaccio: Done! I've added the metrics property as an optional value in the Router.

entrypoint.sh
```diff
@@ -2,4 +2,4 @@
 
 npm run migrate:docker up || exit 1
 npm run seed || exit 1
-npm run start || exit 1
+node ./dist/src/server.js || exit 1
```
nicosantangelo, line 5: are we guaranteed to have the server built in the CI at this point?

package.json
```diff
@@ -8,7 +8,7 @@
     "lint:fix": "tslint -p tsconfig.json --fix",
     "build": "tsc -p tsconfig.json",
     "watch:build": "tsc -p tsconfig.json --watch",
-    "start": "npm run build && node ./dist/src/server.js",
+    "start": "ts-node src/server.ts",
```
nicosantangelo, line 11: Not against the change, but is there any particular reason for it?

src/Forum/Forum.router.ts
```diff
@@ -24,7 +25,7 @@ export class ForumRouter extends Router {
       withAuthentication,
       withCollectionExists,
       withCollectionAuthorization,
-      server.handleRequest(this.post)
+      server.handleRequest(this.post.bind(this))
```
nicosantangelo, line 28: If you wanna avoid using `bind` here you could also use bound instance methods like the Asset.router or Collection.router do:

```ts
// from
async post(req: AuthRequest) {
}

// to
async post = (req: AuthRequest) => {
}
```

src/server.ts
```diff
@@ -65,7 +70,6 @@ if (require.main === module) {
 }
 
 async function startServer() {
-  console.log('Connecting database')
```
nicosantangelo, line 68: Not a terribly useful log, but 😢

nicosantangelo (commented):
Few comments more related to config than the actual code, which looks 💯

entrypoint.sh
LautaroPetaccio: We do! we actually do a `RUN npm run build` on the Dockerfile

package.json
LautaroPetaccio: It's just to have a start command to run the server without the need of recompiling it each time.

src/server.ts
LautaroPetaccio: Ups, I didn't realize I had removed it.

src/Forum/Forum.router.ts
LautaroPetaccio: I used to prefer using variables with methods, but I switched not too long time ago to use bind. I changed it anyways.

nicosantangelo (approved):

PR #251, opened 2021-08-16
feat: Add tos endpoint

LautaroPetaccio:
This PR adds the TOS endpoint to save the user's approval when publishing a collection.


src/warehouse/warehouse.ts
```diff
@@ -0,0 +1,35 @@
+import { env } from 'decentraland-commons'
+
+const WAREHOUSE_URL: string = env.get('WAREHOUSE_URL')
+const WAREHOUSE_TOKEN: string = env.get('WAREHOUSE_TOKEN')
+const WAREHOUSE_CONTEXT_PREFIX: string = env.get('WAREHOUSE_CONTEXT_PREFIX')
```
nicosantangelo, line 5: I think it'd be a good idea to add a throw here explaining that the variables are missing, as they don't have a default value and the ToS eendpoint fails silently https://github.com/decentraland/builder-server/pull/251/files#diff-ea5b87ec142e269f5ad70b6154f928ebe98878ba0656ae127be867ea1a5513d0R114

what do you think?

src/warehouse/warehouse.ts
```diff
@@ -0,0 +1,35 @@
...
+): Promise<void> {
+  const data = {
+    context: WAREHOUSE_CONTEXT_PREFIX
+      ? `${WAREHOUSE_CONTEXT_PREFIX}-${context}`
+      : context,
+    event,
+    body,
+  }
+
+  await fetch(WAREHOUSE_URL, {
```
nicosantangelo, line 27: I think this is working here because the project is importing `isomorphic-fetch` like this: `import 'isomorphic-fetch'` on `graphClient.ts`. Worst still, the project also has a `node-fetch` dependency 🤦 

We don't need to fix this here of course, but maybe we can at least add a `import fetchfrom 'node-fetch'` at the top of the file, to be responsible

nicosantangelo (changes requested):
Tiny comments

src/Project/index.ts
```diff
@@ -1,3 +1,4 @@
 export * from './Project.types'
 export * from './Project.model'
 export * from './Project.router'
+export * from './ProjectByCoord.router'
```
nicosantangelo, line 4: ❤️

src/Collection/Collection.router.ts
```diff
@@ -76,8 +76,8 @@ export class CollectionRouter extends Router {
      */
     this.router.post(
       '/collections/:id/tos',
-      withAuthentication,
-      withCollectionExists,
+      // withAuthentication,
+      // withCollectionExists,
```
nicosantangelo, line 80: Whoops!

src/Collection/Collection.types.ts
```diff
@@ -49,3 +49,17 @@ export const collectionSchema = Object.freeze({
...
+
+export const saveTOSSchema = Object.freeze({
+  type: 'object',
+  properties: {
+    email: {
+      type: 'string',
+      // prettier-ignore
+      pattern: "^[a-zA-Z0-9.!#$%&'*+\/=?^_`{|}~-]+@[a-zA-Z0-9](?:[a-zA-Z0-9-]{0,61}[a-zA-Z0-9])?(?:\.[a-zA-Z0-9](?:[a-zA-Z0-9-]{0,61}[a-zA-Z0-9])?)*$",
+    },
+    collection_address: { type: 'string', pattern: '^0x[a-fA-F0-9]{40}$' },
```
nicosantangelo, line 61: we should be using the custom validators here right?

nicosantangelo (changes requested):

src/Collection/Collection.types.ts
LautaroPetaccio: This can be validated with a regex pattern, we don't really need to build a custom validator for it. I'm assuming that the custom validator will also be a regex, right?

nicosantangelo (approved):

PR #258, opened 2021-08-30
feat: Use redirect for S3 files

LautaroPetaccio:
This PR changes the way we were serving the following endpoints:
- /projects/:id/media/:filename
- /publics/:id/manifest
- /pools/:id/manifest
- /storage/assetPacks/:filename
- /storage/contents/:filename

To respond with a redirect instead of buffering the file before streaming it to the clients.


nicosantangelo (approved):
LGTM

PR #266, opened 2021-09-01
feat: Add cache redirect for other endpoints

LautaroPetaccio:
This PR adds the cache header and the permanently moved status code for the remaining modified endpoints that now redirect to S3.
The redirect to the Amazon S3 bucket is being cached, but the file itself manages its cache trough the bucket headers (etag and last modified).

### DIFF
--- a/src/Manifest/Manifest.router.ts
+++ b/src/Manifest/Manifest.router.ts
@@ -95,8 +95,11 @@ export class ManifestRouter extends Router {
   getProjectManifest = (req: Request, res: Response) => {
     const id = server.extractFromReq(req, 'id')
     const project = new S3Project(id)
+
+    res.setHeader('Cache-Control', 'public,max-age=31536000,immutable')
     return res.redirect(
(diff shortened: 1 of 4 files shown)


src/Manifest/Manifest.router.ts
```diff
@@ -95,8 +95,11 @@ export class ManifestRouter extends Router {
   getProjectManifest = (req: Request, res: Response) => {
     const id = server.extractFromReq(req, 'id')
     const project = new S3Project(id)
+
+    res.setHeader('Cache-Control', 'public,max-age=31536000,immutable')
```
nachomazzara, line 99: What do you think of moving `31536000`  to something like [`DEFAULT_ASSET_PACK_CACHE`](https://github.com/decentraland/builder-server/blob/master/src/AssetPack/AssetPack.router.ts#L28)

nachomazzara: that will work too.

I was thinking also that we can have a new middleware like `withInmutableCache` to be used directly on each router

nachomazzara (approved):

PR #269, opened 2021-09-14
chore: Add logs for getAssets

LautaroPetaccio:
This PR adds logs for the `getAssets` and `getAsset` handlers.
Some functions, like `logExecutionTime` are not abstracted taking into consideration that this is a temporary set of logs to understand what's going on with these endpoints.


nachomazzara (approved):
LGTM

PR #271, opened 2021-09-16
fix: Headers being set after request ended

LautaroPetaccio:


src/Manifest/Manifest.router.ts
```diff
@@ -60,7 +60,7 @@ export class ManifestRouter extends Router {
     this.router.get(
       '/publics/:id/manifest',
       withProjectExists,
-      server.handleRequest(this.getProjectManifest)
+      this.getProjectManifest
```
LautaroPetaccio, line 63: This was using `server.handleRequest` that expects a function that returns a body for the response. It doesn't expect the function to change the headers or to end the request.

src/Share/Share.router.ts
```diff
@@ -38,7 +38,7 @@ export class ShareRouter extends Router {
     const element = await this.findElementByType(id, type)
 
     if (!element) {
-      return res.send(404)
+      return res.status(404).end()
```
LautaroPetaccio, line 41: This was an incorrect send that could potentially hang our service.

Issue #268, opened 2021-09-08
Add curation list


nachomazzara:
- Add a new table `curation` to set collections as ready for review:

```
id: (primary key)
collectionId: uuid (foreign key)
timestamp: Date
```

- Add a new endpoint `@PUT /curations/:collectionId` to set a collection ready for review.

- Add a new endpoint to fetch curations
 * `@GET /curations`: this should return a distinct by collectionId. If it is a member of the committee return all the curations, if it is not, return the curation of your collections. *ALWAYS WITH A DISTINCT BY COLLECTION ID*
* `@GET /curations/:collectionId`: return the last curation per collection id

fzavalia:
In Progress until full integration with builder ui has been tested

PR #279, opened 2021-09-21
feat: Curation table + endpoints

fzavalia:
For Issue #268 

Adds endpoints:

`GET /curations` for obtaining all collections that need review. In the case of being in the committee, all collections are received, if not, only collections belonging to you are.

`GET /curations/:collectionId` To receive the latest curation request for a collection.

`POST /curations/:collectionId` To set a collection as for review


src/Curation/Curation.model.ts
```diff
@@ -0,0 +1,40 @@
+import { Model, raw, SQL } from 'decentraland-server'
+import { CurationAttributes } from './Curation.types'
+
+export class Curation extends Model<CurationAttributes> {
+  static tableName = 'curations'
+
+  static getAll() {
```
fzavalia, line 7: I've based this queries on others I found on the repo, yet I'm kind of worried about interpolating the values directly into the query, which might cause sql injection attacks, (in this case it's not possible because sanitization is done by the router). Do you know if this library supports some kind of sanitization. Eg: `query("WHERE col = $1", [value])`

cazala (approved):

PR #280, opened 2021-09-22
feat: Add asset packs owner query parameter

LautaroPetaccio:
This PR creates two new endpoints:
- The `/assetPacks/default` endpoint, that returns the default asset packs.
- The `/ownAssetPacks` authenticated endpoint, that returns the user's asset packs.

### DIFF
--- a/src/AssetPack/AssetPack.router.ts
+++ b/src/AssetPack/AssetPack.router.ts
@@ -48,6 +48,14 @@ export class AssetPackRouter extends Router {
     const withAssetPackAuthorization = withModelAuthorization(AssetPack)
 
     /**
+     * Get default asset packs
+     */
+    this.router.get(
+      '/assetPacks/default',
+      server.handleRequest(this.getDefaultAssetPacks)
+    )
+
+    /**
      * Get all asset packs
      */
     this.router.get(
@@ -55,6 +63,11 @@ export class AssetPackRouter extends Router {
       withPermissiveAuthentication,
       server.handleRequest(this.getAssetPacks)
     )
+
+    /**
+     * Get asset packs of an address
+     */
+    this.router.get('/ownAssetPacks', withAuthentication, this.getOwnAssetPacks)
 
     /**
      * Get asset pack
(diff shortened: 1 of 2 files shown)


src/AssetPack/AssetPack.router.ts
```diff
@@ -100,22 +100,43 @@ export class AssetPackRouter extends Router {
...
     const ethAddress = req.auth ? req.auth.ethAddress : ''
-    let assetPacks: FullAssetPackAttributes[] = []
+    const owner = req.query.owner
     const tracer = uuidv4()
     this.logger.info(
       `Starting request to get the assets packs with tracer ${tracer} and the address "${ethAddress}"`
     )
 
+    // Process the owner query parameter first
+    if (owner && owner === 'default') {
```
nachomazzara, line 110: I think that `owner` here is redundant.

src/AssetPack/AssetPack.router.ts
```diff
@@ -100,22 +100,43 @@ export class AssetPackRouter extends Router {
...
+      return this.logExecutionTime(
+        this.retrieveDefaultAssetPacks,
+        'Get default asset packs',
+        tracer
+      )
+    } else if (owner && owner === ethAddress) {
```
nachomazzara, line 116: I think that `owner` here is redundant.

src/AssetPack/AssetPack.router.ts
```diff
@@ -100,22 +100,43 @@ export class AssetPackRouter extends Router {
 
   getAssetPacks = async (req: AuthRequest) => {
     const ethAddress = req.auth ? req.auth.ethAddress : ''
-    let assetPacks: FullAssetPackAttributes[] = []
+    const owner = req.query.owner
```
nachomazzara, line 103: I would lowercase `req.query.owner` to sopport checksummed addresses too

nachomazzara: WE can do it in a reutilizable middleware

nachomazzara (changes requested):
LGTM, just minor comments

src/AssetPack/AssetPack.router.ts
LautaroPetaccio: I'll lowercase it!
How would the middleware be? Will it just lowercase an address? Do we need it somewhere else?

src/AssetPack/AssetPack.router.ts
LautaroPetaccio: Your right, I'll remove it.

nachomazzara (approved):
LGTM

PR #286, opened 2021-09-29
feat: Add URN to collection and item

LautaroPetaccio:
This PR does the following:
- Changes the entrypoint to start the server without building it again.
- Adds the URN column to the collection and the items.
- Adds the logic to generate the URNs at the time of retrieving the collections and the items.
- Moves the item schema to another file to fix a circular dependency.

### DIFF
--- a/src/Collection/Collection.router.ts
+++ b/src/Collection/Collection.router.ts
@@ -166,7 +166,9 @@ export class CollectionRouter extends Router {
     // Build the URN for the collections
     return consolidatedCollections.map((collection) => ({
       ...collection,
-      urn: collection.urn ?? getDecentralandCollectionURN(collection),
+      urn:
+        collection.urn ??
+        getDecentralandCollectionURN(collection.contract_address),
     }))
   }
 
@@ -219,7 +221,8 @@ export class CollectionRouter extends Router {
     }
 
     fullCollection.urn =
-      fullCollection.urn ?? getDecentralandCollectionURN(fullCollection)
+      fullCollection.urn ??
+      getDecentralandCollectionURN(fullCollection.contract_address)
 
     return fullCollection
   }
@@ -352,9 +355,15 @@ export class CollectionRouter extends Router {
       data
     )
 
-    // Compute the URN for the Decentraland Collection and set it
-    attributes.urn = getDecentralandCollectionURN(attributes)
-    return new Collection(attributes).upsert()
+    // Sets the DCL collection URN to null before writing it to the DB
+    attributes.urn = null
+    const upsertedCollection = await new Collection(attributes).upsert()
+
+    // Return the collection that was updated or inserted with the DCL URN
+    return {
+      ...upsertedCollection,
+      urn: getDecentralandCollectionURN(upsertedCollection.contract_address),
+    }
   }
 
   deleteCollection = async (req: AuthRequest) => {
--- a/src/Collection/Collection.router.ts
+++ b/src/Collection/Collection.router.ts
@@ -19,10 +19,11 @@ import { Collection, CollectionAttributes } from '../Collection'
 import { isCommitteeMember } from '../Committee'
 import { collectionSchema, saveTOSSchema } from './Collection.types'
 import { RequestParameters } from '../RequestParameters'
-import { hasAccess } from './access'
 import { isPublished } from '../utils/eth'
 import { ItemFragment } from '../ethereum/api/fragments'
 import { sendDataToWarehouse } from '../warehouse'
+import { hasAccess } from './access'
+import { getDecentralandCollectionURN } from './utils'
 
 const validator = getValidator()
 
@@ -158,7 +159,15 @@ export class CollectionRouter extends Router {
       collectionAPI.fetchCollections(),
     ])
 
-    return Bridge.consolidateCollections(dbCollections, remoteCollections)
+    const consolidatedCollections = await Bridge.consolidateCollections(
+      dbCollections,
+      remoteCollections
+    )
+    // Build the URN for the collections
+    return consolidatedCollections.map((collection) => ({
+      ...collection,
+      urn: collection.urn ?? getDecentralandCollectionURN(collection),
+    }))
   }
 
   async getAddressCollections(req: AuthRequest) {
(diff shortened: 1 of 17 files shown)


src/Collection/Collection.types.ts
```diff
@@ -9,16 +15,21 @@ export type CollectionAttributes = {
   minters: string[]
   managers: string[]
   forum_link?: string
-  lock?: Date
-  reviewed_at?: Date
+  lock: Date | null
+  reviewed_at: Date
```
nicosantangelo, line 19: This is correct, it was a mistake on my part because node-pg used to return undefined for NULL values. But while we're at it, you mind fixing the type for reviewed_at and forum_link? both are NULLABLE and are not set at first

src/Collection/utils.ts
```diff
@@ -0,0 +1,32 @@
...
+}
+
+export function toFullCollection(
+  dbCollection: CollectionAttributes
+): FullCollection {
+  const fullCollection: FullCollection = {
+    ...dbCollection,
+    urn: getDecentralandCollectionURN(dbCollection.contract_address),
+  }
+  delete (fullCollection as FullCollection & { urn_suffix: unknown }).urn_suffix
```
nicosantangelo, line 17: we could take advantage of `decentraland-common's` omit here if you like. Avoiding this `delete` for something like
```ts
  const fullCollection: FullCollection = {
    ...utils.omit(dbCollection, ['urn_suffix']),
    urn: getDecentralandCollectionURN(dbCollection.contract_address),
  }
```

or just

```ts
  const { urn_suffix, ...sufixlessDbCollection } = dbCollection
  const fullCollection: FullCollection = {
    ...sufixlessDbCollection,
    urn: getDecentralandCollectionURN(dbCollection.contract_address),
  }
```
but that might upset the linter

src/Collection/utils.ts
```diff
@@ -0,0 +1,32 @@
+import { getCurrentNetworkURNProtocol } from '../ethereum/utils'
+import { CollectionAttributes, FullCollection } from './Collection.types'
+
+export function getDecentralandCollectionURN(
```
nicosantangelo, line 4: Question: You think any of these methods might make sense on the `CollectionService` class? I'm not entirely sure as of right now, but if no, maybe we can think about clear boundaries of these file and the service.

My first intuition is that utils is fine in these cases and if we wanted to extract this, it'd make more sense to do a `CollectionBridge` of sorts with these methods (overkill)

src/Collection/utils.ts
nicosantangelo: Same goes for `toDBCollection` with the urn!

src/Collection/Collection.router.ts
```diff
@@ -146,7 +153,15 @@ export class CollectionRouter extends Router {
...
+      remoteCollections
+    )
+
+    // Build the full collection
+    return consolidatedCollections.map((collection) =>
+      toFullCollection(collection)
+    )
   }
 
   async getAddressCollections(req: AuthRequest) {
```
nicosantangelo, line 167: Maybe we can go the `Item.router.ts` route (heh) here too and add return types to these methods, no?

src/Item/Item.router.ts
```diff
@@ -141,9 +142,11 @@ export class ItemRouter extends Router {
...
       collectionAPI.fetchItems(),
     ])
-    const catalystItems = await peerAPI.fetchWearables(
-      remoteItems.map((item) => item.urn)
-    )
+
+    const catalystItems =
+      remoteItems && remoteItems.length > 0
+        ? await peerAPI.fetchWearables(remoteItems.map((item) => item.urn))
+        : []
```
nicosantangelo, line 149: Why's this necessary? can `remoteItems` be `undefined`? the type says `Promise<ItemFragment[]>`
peer.ts already has this check:
```ts
const wearables: PeerWearable[] =
  urns.length > 0
    ? await this.lambdasClient.fetchWearables({ wearableIds: urns })
    : []
```

same thing [here](https://github.com/decentraland/builder-server/pull/286/files#diff-5db28707b1b7436bf2151ad65251263f4b79eeddc180a1a22996464abd5866c4R181)

src/ethereum/api/Bridge.ts
```diff
@@ -41,13 +41,19 @@ export class Bridge {
...
+    const fullItem = {
       ...dbItem,
+      urn: null,
       in_catalyst: false,
       is_approved: false,
       is_published: false,
       total_supply: 0,
     }
+
+    delete (fullItem as FullItem & { urn_suffix: string | undefined })
```
nicosantangelo, line 53: Maybe we can use utils.omit or destructuring here and [here](https://github.com/decentraland/builder-server/pull/286/files#diff-770f618c954ffa5f7e17dad792dd6e16f45eb3c890e1f0616f18f7bd117af885R166) too?

src/Item/Item.router.ts
LautaroPetaccio: You're right, there was no need to check for undefined. Removed it in both places.

src/Collection/utils.ts
LautaroPetaccio: Done!

src/Collection/Collection.types.ts
LautaroPetaccio: Done!

src/Collection/Collection.router.ts
```diff
@@ -36,6 +41,7 @@ export class CollectionRouter extends Router {
     const withCollectionExists = withModelExists(Collection, 'id')
     const withCollectionAuthorization = withModelAuthorization(Collection)
     const withLowercasedAddress = withLowercasedParams(['address'])
+    console.log('Auth middleware created', withAuthentication)
```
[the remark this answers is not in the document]
LautaroPetaccio: LOL I completely forgot to remove this.

src/Collection/utils.ts
LautaroPetaccio: I'm not entirely sure where this should go, as I think that we would really need more methods like this to have a clear view on how to structure them. The main reason why I'm not sure about moving it into a service is because this method only builds a string that will be used for a collection, but it's not directly involved with collections at all.

Issue #278, opened 2021-09-21
Add `lock` to the collection table


nachomazzara:
- Add a `lock` column to the collection table: TIMESTAMP, NULL.
- Add a new endpoint `@POST /collection/:id/lock` to lock a collection.
- A Decentraland collection is locked if the lock is lower than 1 day from Now and the collection is not in the subgraph (blockchain)
```ts
if (collection.lock + 1 day > Date.now() && collection.isNotPublished()) {
     return true // collection is lock
}
```
- We should reject changes in the collection and items if the collection is locked
- Add tests

PR #289, opened 2021-09-30
feat: check collection lock

nicosantangelo:
Again related to #278 but still missing TPW

### DIFF
--- a/src/Collection/Collection.router.spec.ts
+++ b/src/Collection/Collection.router.spec.ts
@@ -46,10 +46,6 @@ describe('Collection router', () => {
       )
     })
 
-    afterEach(() => {
-      jest.restoreAllMocks()
-    })
-
     it('should update the lock with .now() on the supplied collection id for the owner', async () => {
       return server
         .post(buildURL(url))
--- a/src/Collection/Collection.router.spec.ts
+++ b/src/Collection/Collection.router.spec.ts
@@ -33,6 +33,7 @@ describe('Collection router', () => {
 
   describe('when locking a Collection', () => {
     const now = 1633022119407
+    const url = `/collections/${collectionAttributes.id}/lock`
 
     beforeEach(() => {
       jest.spyOn(Date, 'now').mockReturnValueOnce(now)
@@ -50,8 +51,6 @@ describe('Collection router', () => {
     })
 
     it('should update the lock with .now() on the supplied collection id for the owner', async () => {
-      const url = `/collections/${collectionAttributes.id}/lock`
-
       return server
         .post(buildURL(url))
         .set(createAuthHeaders('post', url))
@@ -63,7 +62,30 @@ describe('Collection router', () => {
             { id: collectionAttributes.id, eth_address: wallet.address }
           )
         })
-        .finally(() => jest.useRealTimers())
+    })
+
+    it('it should fail with an error if the update explodes', async () => {
+      const errorMessage = 'Error message'
+
+      ;(Collection.update as jest.Mock).mockImplementationOnce(() => {
+        throw new Error(errorMessage)
+      })
+
+      return server
+        .post(buildURL(url))
+        .set(createAuthHeaders('post', url))
+        .expect(500)
+        .then(async (response: any) => {
+          expect(response.body).toEqual({
+            ok: false,
+            data: {
+              id: collectionAttributes.id,
+              eth_address: wallet.address,
+              error: errorMessage,
+            },
+            error: "The collection couldn't be updated",
+          })
+        })
     })
   })
 })
--- a/src/Collection/Collection.router.ts
+++ b/src/Collection/Collection.router.ts
@@ -15,12 +15,15 @@ import { peerAPI, Wearable } from '../ethereum/api/peer'
 import { FactoryCollection } from '../ethereum'
 import { Ownable } from '../Ownable'
 import { Item, ItemAttributes } from '../Item'
-import { Collection, CollectionAttributes } from '../Collection'
+import {
+  Collection,
+  CollectionService,
+  CollectionAttributes,
+} from '../Collection'
 import { isCommitteeMember } from '../Committee'
 import { collectionSchema, saveTOSSchema } from './Collection.types'
 import { RequestParameters } from '../RequestParameters'
 import { hasAccess } from './access'
-import { isPublished } from '../utils/eth'
 import { ItemFragment } from '../ethereum/api/fragments'
 import { sendDataToWarehouse } from '../warehouse'
 
@@ -56,7 +59,6 @@ export class CollectionRouter extends Router {
      */
     this.router.get(
       '/collections/:id',
-      withAuthentication,
       withCollectionExists,
       server.handleRequest(this.getCollection)
     )
@@ -299,7 +301,7 @@ export class CollectionRouter extends Router {
     } catch (error) {
       throw new HTTPError(
         "The collection couldn't be updated",
-        { id, eth_address, error },
+        { id, eth_address, error: (error as Error).message },
         STATUS_CODES.error
       )
     }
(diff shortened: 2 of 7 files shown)


src/Collection/Collection.router.spec.ts
```diff
@@ -63,7 +58,30 @@ describe('Collection router', () => {
             { id: collectionAttributes.id, eth_address: wallet.address }
           )
         })
-        .finally(() => jest.useRealTimers())
+    })
+
+    it('it should fail with an error if the update explodes', async () => {
```
LautaroPetaccio, line 63: What if we wrap this `it` in a `describe` to describe the context and the use the `beforeEach` to code the context:
```
describe('when the lock update fails', () => {
  beforeEach(() => {
      ;(Collection.update as jest.Mock).mockImplementationOnce(() => {
        throw new Error(errorMessage)
      })
  })
})
```
We can later have the `it` describe only what is expected from the context and the execution of the request:
```
it('it should respond with a 500 and and error message saying that the collection couldn't be updated'
```

We can also do the same for the other case, specifying that the correct response will occur whenever the update works.

LautaroPetaccio: Beware that you have an async closure here but there's no await.

src/Collection/Collection.router.spec.ts
```diff
@@ -63,7 +58,30 @@ describe('Collection router', () => {
...
+
+      ;(Collection.update as jest.Mock).mockImplementationOnce(() => {
+        throw new Error(errorMessage)
+      })
+
+      return server
+        .post(buildURL(url))
+        .set(createAuthHeaders('post', url))
+        .expect(500)
+        .then(async (response: any) => {
```
LautaroPetaccio, line 74: This async definition can be removed.

src/Collection/Collection.service.spec.ts
```diff
@@ -0,0 +1,92 @@
+import { v4 as uuidv4 } from 'uuid'
+import { Collection } from './Collection.model'
+import { CollectionService } from './Collection.service'
+import { wallet } from '../../spec/utils'
+
+jest.mock('./Collection.model')
+
+describe.only('Collection service', () => {
```
LautaroPetaccio, line 8: It's only meeeee.

src/Collection/Collection.router.ts
```diff
@@ -348,9 +350,11 @@ export class CollectionRouter extends Router {
       )
     }
 
-    if (await this.isCollectionPublished(id)) {
+    const service = new CollectionService()
```
LautaroPetaccio, line 353: The collection service is stateless, what do you think about having it as an instance variable?

src/Collection/Collection.router.ts
```diff
@@ -375,10 +379,11 @@ export class CollectionRouter extends Router {
...
     const id = server.extractFromReq(req, 'id')
+    const service = new CollectionService()
 
-    if (await this.isCollectionPublished(id)) {
+    if (await service.isLocked(id)) {
       throw new HTTPError(
-        "The collection is published. It can't be deleted",
+        `Cannot access the Collection ${id} because it's either locked or published`,
         { id },
         STATUS_CODES.unauthorized
```
LautaroPetaccio, line 388: I wouldn't mix both errors, as it can be used to represent two different errors that we would like to show the user.
Also, the collection locked error (and maybe the error related to the collection being already published) should return a 423 (locked) instead of a 401 that relates to a user not being authenticated.

src/Collection/Collection.service.spec.ts
```diff
@@ -0,0 +1,92 @@
...
+
+    it('should return false if the collection does not exist', async () => {
+      findOneMock.mockImplementationOnce((collectionId: string) => {
+        if (collectionAttributes.id === collectionId) {
+          return undefined
+        }
+        throw new Error('Invalid id')
+      })
+      expect(await service.isLocked(collectionAttributes.id)).toBe(false)
+    })
```
LautaroPetaccio, line 41: What do you think about using a `describe` here to describe the context and a `beforeEach` to code it?
```suggestion
describe('when the collection doesn't exist', () => {
    beforeEach(() => {
      findOneMock.mockImplementationOnce((collectionId: string) => {
        if (collectionAttributes.id === collectionId) {
          return undefined
        }
        throw new Error('Invalid id')
      })
    })
    it('should return false', async () => {
      expect(await service.isLocked(collectionAttributes.id)).toBe(false)
    })
})
```

src/Collection/Collection.service.spec.ts
```diff
@@ -0,0 +1,92 @@
...
+      managers: [],
+      forum_link: null,
+      lock: null,
+      reviewed_at: new Date(),
+      created_at: new Date(),
+      updated_at: new Date(),
+    }
+
+    const service = new CollectionService()
+    const findOneMock = Collection.findOne as jest.Mock
```
LautaroPetaccio, line 28: What if we do a mockRestore of `findOneMock` on a `beforeEach` to be sure that it gets cleaned after every execution?

LautaroPetaccio (changes requested):

src/Collection/Collection.router.ts
nicosantangelo: I like it! will change

src/Collection/Collection.service.spec.ts
nicosantangelo: I like it! I'm not really used to use describe to explain the context in these situations but I like how it looks/structures the code 👍

src/Collection/Collection.router.spec.ts
```diff
@@ -45,25 +46,52 @@ describe('Collection router', () => {
...
-        .set(createAuthHeaders('post', url))
-        .expect(200)
-        .then(async (response: any) => {
-          expect(response.body).toEqual({ data: true, ok: true })
-          expect(Collection.update).toHaveBeenCalledWith(
-            { lock: new Date(now) },
-            { id: collectionAttributes.id, eth_address: wallet.address }
-          )
+      beforeEach(() => {
+        ;(Collection.update as jest.Mock).mockImplementationOnce(() => {
```
LautaroPetaccio, line 73: As `Collection.update` returns a promise, I would use `.mockRejectedValueOnce(new Error(errorMessage))`.

src/Collection/Collection.router.ts
```diff
@@ -376,42 +390,28 @@ export class CollectionRouter extends Router {
...
       )
     }
 
+    if (this.service.isLockExpired(collection.lock)) {
+      throw new HTTPError(
+        "The collection is locked. It can't be deleted",
+        { id },
+        STATUS_CODES.unauthorized
```
LautaroPetaccio, line 407: What do you think about returning a 423 status code (Locked)? Unauthorized refers to the authentication of the user.

src/Collection/Collection.service.spec.ts
```diff
@@ -0,0 +1,29 @@
...
+
+      it('should return false if the lock date + 1 day is newer than now', async () => {
+        jest
+          .spyOn(Date, 'now')
+          .mockReturnValueOnce(Date.now() + twoDaysInMilliseconds)
+
+        expect(await service.isLockExpired(lock)).toBe(false)
+      })
+    })
+  })
```
LautaroPetaccio, line 28: This is how it should look if we follow the describe and beforeEach approach:
```suggestion
    describe('when the collection is locked with a date older than a day', () => {
      let lock;
      beforeEach(() => {
        lock = new Date()
        jest
          .spyOn(Date, 'now')
          .mockReturnValueOnce(Date.now() + twoDaysInMilliseconds)
      })
      afterAll(() => {
       // Just in case something goes wrong with the test and the Date.now function never gets executed.
        jest.restoreAllMocks()
      })

      it('should return false', async () => {
        expect(await service.isLockExpired(lock)).toBe(false)
      })
    })
  })
```

src/Collection/Collection.service.spec.ts
```diff
@@ -0,0 +1,29 @@
...
+})
```
LautaroPetaccio, line 29: You're lacking the positive test, the one that tests when a collection is locked.

LautaroPetaccio (changes requested):

src/Collection/Collection.router.ts
nicosantangelo: Much better 👍

LautaroPetaccio (approved):

PR #290, opened 2021-10-01
fix: Add async and default error handler

LautaroPetaccio:
This PR adds an async handler wrapper that catches errors in async handlers and calls the next function so the errors can be  handled by the express error handlers. A simple error handler is added too, replicating the error handler in the decentraland server code.

### DIFF
--- a/src/common/ErrorHandling.spec.ts
+++ b/src/common/ErrorHandling.spec.ts
@@ -1,0 +1,107 @@
+import supertest from 'supertest'
+import express from 'express'
+import { buildURL } from '../../spec/utils'
+import { app } from '../server'
+import { HTTPError } from './HTTPError'
+import { asyncHandler } from './asyncHandler'
+
+const errorMessage = 'anErrorMessage'
+const errorStatusCode = 400
+const errorData = { id: 'anId' }
+
+const handlerWithHttpError = () => {
+  throw new HTTPError(errorMessage, errorData, errorStatusCode)
+}
+const handlerWithCommonError = () => {
+  throw new HTTPError(errorMessage, errorData, errorStatusCode)
+}
+const handlerWithSentHeadersAndCommonError = (
+  _: express.Request,
+  res: express.Response
+) => {
+  res.sendStatus(401)
+  throw new HTTPError(errorMessage, errorData, errorStatusCode)
+}
+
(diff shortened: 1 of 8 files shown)


src/common/ErrorHandling.spec.ts
```diff
@@ -0,0 +1,107 @@
...
+import { asyncHandler } from './asyncHandler'
+
+const errorMessage = 'anErrorMessage'
+const errorStatusCode = 400
+const errorData = { id: 'anId' }
+
+const handlerWithHttpError = () => {
+  throw new HTTPError(errorMessage, errorData, errorStatusCode)
+}
+const handlerWithCommonError = () => {
```
fzavalia, line 15: Are both handlerWithHttpError and handlerWithCommonError supposed to throw an HTTPError?

LautaroPetaccio: They're not! I'll fix those issues

LautaroPetaccio: Done, fixed.

src/common/ErrorHandling.spec.ts
```diff
@@ -0,0 +1,107 @@
...
+          ok: false,
+        })
+      })
+  })
+})
+
+describe('when handling sync http errors', () => {
+  it('should respond with the a 500 an the error message', () => {
+    return server
+      .get(buildURL('/syncError'))
```
fzavalia, line 82: This endpoint does not throw an http error, just a common error (testing the same as the next test)

src/common/ExpressApp.ts
```diff
@@ -118,7 +118,7 @@ export class ExpressApp {
     return this
   }
 
-  use(...handers: express.RequestHandler[]) {
+  use(...handers: express.RequestHandler[] | express.ErrorRequestHandler[]) {
```
fzavalia, line 121: typo in handers -> handlers

fzavalia (approved):
LGTM!

src/common/ErrorHandling.spec.ts
LautaroPetaccio: Changed to syncHttpError!

src/common/ExpressApp.ts
LautaroPetaccio: Fixed it!

PR #296, opened 2021-10-06
fix: Upsert item without converting it to a DB item

LautaroPetaccio:
This PR does the following:
- Moves the item schema to a different file, to fix a circular dependency.
- Creates a `toDBItem` to convert a `FullItem` to a `ItemAttributes` so inserting those items works correctly.

### DIFF
--- a/src/Item/Item.router.ts
+++ b/src/Item/Item.router.ts
@@ -368,7 +368,7 @@ export class ItemRouter extends Router {
     const attributes = toDBItem({
       ...itemJSON,
       eth_address,
-    }) as ItemAttributes
+    })
 
     if (id !== attributes.id) {
       throw new HTTPError('The body and URL item ids do not match', {
--- a/src/Item/Item.schema.ts
+++ b/src/Item/Item.schema.ts
@@ -2,6 +2,7 @@ import { metricsSchema } from '../Metrics/Metrics.types'
 import { ItemRarity, ItemType } from './Item.types'
 import { wearableSchema } from './wearable/types'
 
+// The schema is placed into this file to avoid a circular dependency.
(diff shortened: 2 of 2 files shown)


src/Item/Item.router.ts
```diff
@@ -363,10 +365,10 @@ export class ItemRouter extends Router {
       }
     }
 
-    const attributes = {
+    const attributes = toDBItem({
       ...itemJSON,
       eth_address,
-    } as ItemAttributes
+    }) as ItemAttributes
```
cazala, line 371: there's no need for the `as ItemAttributes` anymore right?

cazala (approved):
LGTM!

src/Item/Item.schema.ts
```diff
@@ -0,0 +1,47 @@
+import { metricsSchema } from '../Metrics/Metrics.types'
+import { ItemRarity, ItemType } from './Item.types'
+import { wearableSchema } from './wearable/types'
+
+export const itemSchema = Object.freeze({
```
nicosantangelo, line 5: Question: Why are we splitting this schema into its own file here but not in the other models?

nicosantangelo: I'm not opposed to that, mostly if it's a circular dep concern!
Still, I think it's valid to say that an schema can live in `XX.types.ts` if we look at that file as the description of how looks. Should we add an issue so we can track it?

LautaroPetaccio: Done! https://app.zenhub.com/workspaces/dapps-5ffc44ecec9f8500140e173c/issues/decentraland/builder-server/298

PR #300, opened 2021-10-06
fix: Bridged items not being added correctly

LautaroPetaccio:
This PR fixes an issue in the item's bridging that occurred when the items weren't published.


nicosantangelo (approved):
LGTM

PR #305, opened 2021-10-12
feat: Add support for the creation of Third Party Collections

LautaroPetaccio:
This PR does the following:
- Splits the code between inserting a Third Party Collection and a Decentraland Collection so it's clearer to follow.
- Moves the code of upserting a collection into the service.
- Tests the code to upsert a collection.
- Adds a temporal `isManager` method that checks if the user's address is one of the ones being configured to allow the insertion of TPC.

### DIFF
--- a/src/Collection/Collection.router.ts
+++ b/src/Collection/Collection.router.ts
@@ -335,14 +335,10 @@ export class CollectionRouter extends Router {
 
   upsertCollection = async (req: AuthRequest): Promise<FullCollection> => {
     const id = server.extractFromReq(req, 'id')
-    let collectionJSON: FullCollection
-    try {
-      collectionJSON = server.extractFromReq(req, 'collection')
-    } catch (error) {
-      throw new HTTPError((error as Error).message).setStatusCode(
-        STATUS_CODES.badRequest
-      )
-    }
+    const collectionJSON: FullCollection = server.extractFromReq(
+      req,
+      'collection'
+    )
 
     const eth_address = req.auth.ethAddress
 


.ci/index.ts
```diff
@@ -134,6 +134,11 @@ export = async function main() {
         name: 'WAREHOUSE_TOKEN',
         value: config.requireSecret('WAREHOUSE_TOKEN'),
       },
+      {
+        name: 'TPW_MANAGER_ADDRESSES',
+        value:
+          '0x1D9aa2025b67f0F21d1603ce521bda7869098f8a,0xeDaE96F7739aF8A7fB16E2a888C1E578E1328299',
```
nicosantangelo, line 140: maybe we could add @fzavalia and @cazala addresses here too

src/Collection/Collection.model.ts
```diff
@@ -19,11 +19,31 @@ export class Collection extends Model<CollectionAttributes> {
...
+  ): Promise<boolean> {
     const counts = await this.query(SQL`
     SELECT count(*) as count
       FROM ${raw(this.tableName)}
       WHERE id != ${id}
+        AND (urn_suffix IS NULL${
+          urn_suffix
+            ? ` OR LOWER(urn_suffix) = ${urn_suffix.toLowerCase()}`
+            : ''
+        })
```
nicosantangelo, line 46: This shouldn't really work! the `SQL` function takes a template, grabs each interpolated value and constructs a valid node-psql query like:

```ts
const id = 3
const name = 'hey'
const a = SQL`
  SELECT * FROM collections WHERE id = ${id} AND name = ${name}`

a // => { query: 'SELECT * FROM collections where id = $1', values: [3, 'hey'] }
```

To append arbitrary data we use `raw` (or `SQL.raw`), but we have to be careful no to do it with data we do not control. `this.tableName` is fine, `urn_suffix` might not be.

This query would then result in something like:

```sql
    SELECT count(*) as count
      FROM collections
      WHERE id != $1
        AND (urn_suffix IS NULL$2)
        AND LOWER(name) = $3
```

Breaking with ` syntax error at or near \"NULL$2\"`

You could instead append `SQL` functions, as they support that kind of behaviour:

```ts
    const a = SQL`
    SELECT count(*) as count
      FROM ${raw(this.tableName)}
      WHERE id != ${id}
        AND (urn_suffix IS NULL ${
          urn_suffix
            ? SQL`OR LOWER(urn_suffix) = ${urn_suffix.toLowerCase()}`
            : SQL`OR 1 = 1`
        })
        AND LOWER(name) = ${name.toLowerCase()}`

// Resulting in:
`SELECT count(*) as count
      FROM collections
      WHERE id != $1
        AND (urn_suffix IS NULL OR LOWER(urn_suffix) = $2)
        AND LOWER(name) = $3`
        
// or

`SELECT count(*) as count
      FROM collections
      WHERE id != $1
        AND (urn_suffix IS NULL OR 1 = 1)
        AND LOWER(name) = $2`
```

src/Collection/Collection.router.ts
```diff
@@ -337,83 +335,48 @@ export class CollectionRouter extends Router {
...
-        STATUS_CODES.unauthorized
+        'Invalid schema',
+        validate.errors,
+        STATUS_CODES.badRequest
       )
     }
 
-    const canUpsert = await new Ownable(Collection).canUpsert(id, eth_address)
-    if (!canUpsert) {
+    // Is it ok if we have it here before the other checks?
```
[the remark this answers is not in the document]
nicosantangelo: I am not sure if this check applies to TPW collections or not, but the check it's fine here for DCL collections

src/Collection/Collection.service.ts
```diff
@@ -1,16 +1,129 @@
...
+const DAY = 1000 * 60 * 60 * 24
 export class CollectionService {
   isLockActive(lock: Date | null) {
     if (!lock) {
       return false
     }
 
     const deadline = new Date(lock)
-    deadline.setDate(deadline.getDate() + 1)
+    return deadline.getTime() + DAY > Date.now()
```
nicosantangelo, line 19: If we're not using `get/setDate()` anymore, we can just add `DAY` to the lock value:

```suggestion
    return lock + DAY > Date.now()
```

src/Collection/Collection.service.ts
```diff
@@ -1,16 +1,129 @@
...
+  }
+
+  private async checkIfNameIsValid(
+    id: string,
+    name: string,
+    urn_suffix?: string
+  ): Promise<void> {
+    if (!(await Collection.isValidName(id, name.trim(), urn_suffix))) {
+      throw new HTTPError(
```
nicosantangelo, line 28: To discuss: For both this method and the ones below. Throwing `HTTPError`s seems a bit of a implementation detail right?
I don't want to make the code over-complex without needing to, but we just coupled some of the `CollectionService` to the Router, as it wouldn't make sense for this error to be thrown anywhere else.

So for example `checkIfNameIsValid` looks like a router method, as is just using `Collection.isValidName` and throwing an `HTTPError`

I understand why this is here, and the obvious solution would be to throw a more generalized error and then catching on the router and re-throwing an HTTPError (which might be hard for STATUS_CODES)

So I don't mean to change the implementation but wanted to know what you though of this and if you had it in mind when writing the code

src/Collection/Collection.service.ts
```diff
@@ -1,16 +1,129 @@
...
+    const [, urn_suffix] = decodeTPCollectionURN(collectionJSON.urn)
+    if (!(await isTPWManger(collectionJSON.urn, eth_address))) {
+      throw new HTTPError(
+        'Unauthorized',
+        { id, urn: collectionJSON.urn, eth_address },
+        STATUS_CODES.unauthorized
+      )
+    }
+
+    // should names be shared with the other collections?
```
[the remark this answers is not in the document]
nicosantangelo: I'm not sure really, but I don't think we have to
cc: @nachomazzara

src/Collection/Collection.service.ts
```diff
@@ -1,16 +1,129 @@
...
+          "The collection is locked. It can't be saved",
+          { id },
+          STATUS_CODES.locked
+        )
+      }
+    }
 
-    return deadline.getTime() > Date.now()
+    const attributes = toDBCollection(collectionJSON)
+    // Should we do something with the salt and the contract address? There's no need to have them
```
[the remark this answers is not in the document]
nicosantangelo: Both `salt` and `contract_address` should be `NULLable` properties right? I think the `CollectionAttributes` does not reflect this correctly (I think we should fix that), but the table does not have a `NOT NULL` so we can just leave them empty right?

src/Collection/utils.ts
```diff
@@ -2,26 +2,56 @@ import { utils } from 'decentraland-commons'
...
+    contract_address,
+    salt,
+  }
+}
+
+export function isTPCollection(urn: string) {
+  return tpwCollectionURNRegex.test(urn)
+}
+
+export function decodeTPCollectionURN(urn: string): [string, string] {
```
nicosantangelo, line 50: I think it'd be useful to return an object with names for each value. I found `decodeTPCollectionURN` in the code above and couldn't really tell what the response value was. Reading the implementation I can't really tell either

What do you think?

src/ethereum/api/tpw.ts
```diff
@@ -0,0 +1,14 @@
+import { env } from 'decentraland-commons'
+const MANAGERS = env.get('TPW_MANAGER_ADDRESSES', '').split(/[ ,]+/)
+/**
+ * Checks if an address manages a third party wearable collection.
+ *
+ * @param urn - The URN of the TWP collection where to get the information about the collection.
+ * @param address - The address to check if it manages the collection.
+ */
+export function isManager(_: string, address: string): Promise<boolean> {
+  if (MANAGERS.includes(address)) {
```
nicosantangelo, line 10: I know this is a mock implementation for now, but careful with mix cases. We should probably lowercase everything

nicosantangelo (commented):
Good Work!

.ci/index.ts
```diff
@@ -134,6 +134,11 @@ export = async function main() {
         name: 'WAREHOUSE_TOKEN',
         value: config.requireSecret('WAREHOUSE_TOKEN'),
       },
+      {
+        name: 'TPW_MANAGER_ADDRESSES',
+        value:
+          '0x1D9aa2025b67f0F21d1603ce521bda7869098f8a,0xeDaE96F7739aF8A7fB16E2a888C1E578E1328299,0xeC6E6c0841a2bA474E92Bf42BaF76bFe80e8657C,0x24e5F44999c151f08609F8e27b2238c773C4D020',
```
nicosantangelo, line 140: Doxxed

src/Collection/Collection.router.ts
```diff
@@ -108,7 +110,7 @@ export class CollectionRouter extends Router {
      */
     this.router.put(
       '/collections/:id',
-      withAuthentication,
+      // withAuthentication,
```
nicosantangelo, line 113: whops!

src/Collection/Collection.router.ts
```diff
@@ -333,96 +335,92 @@ export class CollectionRouter extends Router {
...
+    let collectionJSON: FullCollection
+    try {
+      collectionJSON = server.extractFromReq(req, 'collection')
+    } catch (error) {
+      throw new HTTPError(
+        (error as Error).message,
+        null,
+        STATUS_CODES.badRequest
+      )
+    }
```
nicosantangelo, line 347: Are we doing this to avoid a 500 error? I ask because the very same thing is happening with all other `server.extractFromReq` calls. Say `server.extractFromReq(req, 'data')` below for example

also you can avoid the null by doing `new HTTPError(error.message).setStatusCode(STATUS_CODES.badRequest)` as it returns `this`

nicosantangelo (commented):
what an expressive piece of code `!!!!!`

I think it looks great, just a couple of comments

nicosantangelo (approved):

LautaroPetaccio:
Closed #308

PR #306, opened 2021-10-12
feat: Implement middleware for schema validations

LautaroPetaccio:
This PR implements a middleware to do request body validations.
Each request is validated through a request schema and, in case of errors, the response is formatted as it was done before, with the exception of the status code that now is a 400 (bad request).


fzavalia (approved):
LGTM! question is just out of curiosity

PR #307, opened 2021-10-13
feat: return third party records

### LINKED ISSUE #275: Fetch Third parties
- Add a new endpoint to get the third-parties from the third-party subgraph.
- Add a new type `ThirdParty`.
- Util method to check if a specific Ethereum Address is manager from a third-party.
- Util method to check if a third party exists. (by URN)

```
{
  id: "urn:decentraland:polygon:ext-thirdparty",
  managers: [0xNacho, 0xAlvaro, ..., 0xBiza],
  maxItems: 0,
  items: [],
  name: "name",
  description: "desc",
  collections: [colllection_id1, collection_id2,...],
  resolver: "https://api.thirdparty/v1/",
  isApproved: true
}
```

nicosantangelo:
Closes #275

### DIFF
--- a/spec/utils.ts
+++ b/spec/utils.ts
@@ -37,10 +37,12 @@ export const wallet: Wallet = {
   },
 }
 
-export function buildURL(uri: string, queryString?: Record<string, string>) {
+export function buildURL(
+  uri: string,
+  queryString: Record<string, string> = {}
+) {
   const API_VERSION = env.get('API_VERSION', 'v1')
-  const search = queryString ? buildSearch(queryString) : ''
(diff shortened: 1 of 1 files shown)


spec/utils.ts
```diff
@@ -37,18 +37,33 @@ export const wallet: Wallet = {
   },
 }
 
-export function buildURL(uri: string) {
+export function buildURL(uri: string, queryString?: Record<string, string>) {
```
LautaroPetaccio, line 40: What if we do the same as in `createAuthHeaders` with the default parameter? You'll be able to return the string plus the query parameters.

PR #310, opened 2021-10-14
feat: only return approved third parties

nicosantangelo:
It also removed __typename from requests as we're not using it and it's cluttering our server responses (from thegraph)

### DIFF
--- a/src/ThirdParty/utils.spec.ts
+++ b/src/ThirdParty/utils.spec.ts
@@ -1,0 +1,38 @@
+import {
+  ThirdPartyFragment,
+  ThirdPartyMetadataType,
+} from '../ethereum/api/fragments'
+import { toThirdParty } from './utils'
+import { ThirdParty } from './ThirdParty.types'
+
+describe('toThirdParty', () => {
+  describe('when a fragment is supplied', () => {
+    let fragment: ThirdPartyFragment
+
+    beforeEach(() => {
+      fragment = {
+        id: 'some:id',
+        managers: ['0x1', '0x2'],
+        metadata: {
+          type: ThirdPartyMetadataType.THIRD_PARTY_V1,
+          thirdParty: {
+            name: 'a name',
+            description: 'some description',
+          },
+        },
+      }
+    })
+
+    it('should take a third party fragment and parse it to conform to the ThirdParty type', async () => {
+      const { name, description } = fragment.metadata.thirdParty
+
+      const thirdParty: ThirdParty = {
+        id: fragment.id,
+        managers: fragment.managers,
+        name: name,
+        description: description,
+      }
+      expect(toThirdParty(fragment)).toEqual(thirdParty)
+    })
+  })
+})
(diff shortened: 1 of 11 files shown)


src/ethereum/api/fragments.ts
```diff
@@ -53,9 +53,6 @@ export const thirdPartyFragment = () => gql`
   fragment thirdPartyFragment on ThirdParty {
     id
     managers
-    isApproved
-    maxItems
-    totalItems
```
LautaroPetaccio, line 58: Unless we're creating another endpoint to get the max items and the total items, I would leave these properties here so we can use them in the UI to tell the user to buy more slots.

LautaroPetaccio (changes requested):

src/ThirdParty/utils.ts
```diff
@@ -0,0 +1,12 @@
+import { utils } from 'decentraland-commons'
+import { ThirdPartyFragment } from '../ethereum/api/fragments'
+import { ThirdParty } from './ThirdParty.types'
+
+export function toThirdParty(fragment: ThirdPartyFragment): ThirdParty {
+  const { name, description } = fragment.metadata.thirdParty
```
LautaroPetaccio, line 6: I wouldn't use the object deconstruction here as thirdParty can be null.

src/ethereum/api/fragments.ts
```diff
@@ -106,9 +112,21 @@ export type CollectionFragment = {
...
   totalItems: number
+  metadata: ThirdPartyMetadata
+}
+
+export type ThirdPartyMetadata = {
+  type: ThirdPartyMetadataType
+  thirdParty: {
+    name: string
+    description: string
+  }
```
LautaroPetaccio, line 125: The property thirdParty can be null

src/ThirdParty/utils.spec.ts
```diff
@@ -0,0 +1,42 @@
...
+          type: ThirdPartyMetadataType.THIRD_PARTY_V1,
+          thirdParty: {
+            name: 'a name',
+            description: 'some description',
+          },
+        },
+      }
+    })
+
+    it('should take a third party fragment and parse it to conform to the ThirdParty type', async () => {
```
LautaroPetaccio, line 28: We can remove the async here, as we're not using the await keyword:
```suggestion
    it('should take a third party fragment and parse it to conform to the ThirdParty type', () => {
```

src/middleware/authentication.ts
```diff
@@ -74,7 +74,7 @@ async function decodeAuthChain(req: Request): Promise<string> {
       errorMessage = 'Missing ETH address in auth chain'
     } else {
       try {
-        const endpoint = (req.method + ':' + req.url).toLowerCase()
+        const endpoint = (req.method + ':' + req.path).toLowerCase()
```
LautaroPetaccio, line 77: Do we have any kind of test that verifies this change?

LautaroPetaccio (changes requested):

src/middleware/authentication.ts
nicosantangelo: We don't actually but I felt like testing the entire authentication middleware was a bit overkill for this PR

PR #313, opened 2021-10-15
fix: allow for published item updates even when the collection is locked

nicosantangelo:
Closes #311

PR #314, opened 2021-10-15
refactor: use collection service published

nicosantangelo:
depends on https://github.com/decentraland/builder-server/pull/313

The tests are fixed on https://github.com/decentraland/builder-server/pull/310 and broken in master also

### DIFF
--- a/src/Item/Item.router.ts
+++ b/src/Item/Item.router.ts
@@ -378,7 +378,8 @@ export class ItemRouter extends Router {
     }
 
     const isDbCollectionPublished =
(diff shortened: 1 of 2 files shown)


src/Item/Item.router.ts
```diff
@@ -347,30 +347,11 @@ export class ItemRouter extends Router {
       }
     }
 
+    const service = new CollectionService()
```
LautaroPetaccio, line 350: Can we init the `CollectionService` globaly?

PR #321, opened 2021-10-21
fix: Unordered published collections

LautaroPetaccio:
This PR fixed an issue that assumed that the remote items would be ordered, resulting in the db items having a different blockchain item.

### DIFF
--- a/src/Collection/Collection.router.ts
+++ b/src/Collection/Collection.router.ts
@@ -11,26 +11,26 @@ import {
 } from '../middleware'
 import { collectionAPI } from '../ethereum/api/collection'
 import { Bridge } from '../ethereum/api/Bridge'
+import { ItemFragment } from '../ethereum/api/fragments'
 import { peerAPI, Wearable } from '../ethereum/api/peer'
 import { FullItem, Item, ItemAttributes } from '../Item'
+import { isCommitteeMember } from '../Committee'
+import { RequestParameters } from '../RequestParameters'
+import { sendDataToWarehouse } from '../warehouse'
+import { Collection } from './Collection.model'
 import {
-  Collection,
   CollectionService,
-  CollectionAttributes,
   CollectionLockedException,
   CollectionAlreadyPublishedException,
   WrongCollectionException,
   UnauthorizedCollectionEditException,
-} from '../Collection'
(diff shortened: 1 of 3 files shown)


src/Collection/Collection.router.ts
```diff
@@ -263,22 +263,30 @@ export class CollectionRouter extends Router {
...
-      const updates = []
 
       remoteItems = fetches[0]
       catalystItems = fetches[1]
 
-      for (const [index, remoteItem] of remoteItems.entries()) {
-        updates.push(
-          Item.update(
+      await Promise.all(
+        items.map((item, index) => {
```
nachomazzara, line 271: I know that functional programming sometimes looks pretty and fachera but IMO in this critical case I would prefer to read something more imperative

LautaroPetaccio: Are you agains the map or the concept of having the promise all surrounding the map?
The map differentiates from the for in that it removes the need of having to do the last `updates.push`, removing the need to understand that there's an array outside of the scope that can be modified and that could contain other items.
I can put everything into a `updates` variables as before and then do a `Promise.all`, or change it for a for, what do you think that it's best?

src/Collection/Collection.router.ts
```diff
@@ -263,22 +263,30 @@ export class CollectionRouter extends Router {
         ),
         peerAPI.fetchWearables(remoteItems.map((item) => item.urn)),
```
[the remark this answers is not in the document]
LautaroPetaccio: I fixed it and requested the wearables after the updates.

spec/mocks/items.ts
```diff
@@ -102,3 +102,16 @@ export const itemFragmentMock = {
   metadata: {},
   contentHash: '',
 }
+
+export function convertItemDatesToISO<T extends ItemAttributes | FullItem>(
```
nicosantangelo, line 106: This is called convertItemDatesToISO but it takes and uses a collection. Is that an naming error?
an item does not have a `reviewed_at`

src/Collection/Collection.router.ts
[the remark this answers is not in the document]
nicosantangelo: Probably a copypaste, don't think we did it on purpose tbh

PR #327, opened 2021-10-22
feat: Add support for TPC deletion

### LINKED ISSUE #316: Add support for deleting Third Party Collection
Third Party Collections should be deleted only if they don't contain a published item.

LautaroPetaccio:
This PR adds the needed checks for the Third Party Collections.
The PR includes the following:
- Adds a method to the `withModelAuthorization` middleware so that the model authorization can be customized.
- Adds a custom check to the `withModelAuthorization` to do third party checks for the collections and the items.
- Adds multiple mocking utilities to reduce the mocking code.
- Adds the code to get third party items from the graph.
- Adds the code to delete third party collections and its tests.
- Adds tests to verify that the modified middleware works (the item tests test a part of the item deletion, verifying that the middleware works for non third party items, that is, that we didn't break anything).
- Moves the wallet mock to a new file to solve a circular dependency.
- Removes unused mocks in the collections and items tests.
- Adds the THIRD_PARTY_URL environment variable so dev works with mumbai.
Closes #316


src/Collection/Collection.service.ts
```diff
@@ -1,21 +1,31 @@
...
 import { Collection } from './Collection.model'
 
+enum CollectionType {
+  THIRD_PARTY,
+  DCL,
+}
 export class CollectionLockedException extends Error {
-  constructor(public id: string) {
-    super("The collection is locked. It can't be saved.")
+  constructor(public id: string, action: string) {
```
nicosantangelo, line 17: I'm not entirely sold on the idea of having a generic string to serve as an error message. I think is a lot to put on the caller to know which action is being performed and to know exactly where the verb/word is going to end up in the error message.
What do you think of:
- Using an enum for each supported action
- Adding an error for each supported action which internally uses the same concept (just not exposed to the public API)

In conjunction to all of this, maybe we can move these errors to their own file?

src/Collection/Collection.service.ts
```diff
@@ -131,4 +151,70 @@ export class CollectionService {
 
     return true
   }
+
+  public async deleteCollection(collectionId: string): Promise<void> {
+    const collection = await this.getDBCollection(collectionId)
+    if (collection.urn_suffix !== null && collection.third_party_id !== null) {
```
nicosantangelo, line 157: We might want to at least add a TODO here, as this check is repeated in `toFullCollection` and on line 195 here. I think it's a good candidate for a `" is third party "` method. In [this PR](https://github.com/decentraland/builder-server/pull/335/files) I renamed the current isTPWCollection to show that the method was using the URN to check, so we have the name available

src/Collection/Collection.service.ts
```diff
@@ -131,4 +151,70 @@ export class CollectionService {
...
+        ethAddress
+      )
+    } else if (collection) {
+      return collection.eth_address === ethAddress
+    }
+
+    return false
+  }
+
+  private async getDBCollection(
```
nicosantangelo, line 210: Questions:
- We have a `getDBCollection` on the utils file that just does this: `export const getDBCollection = (id: string) => Collection.findOne<CollectionAttributes>(id)` should that method be updated to throw and used here too?
- should we use this method in the rest of the file?

src/Item/Item.service.ts
```diff
@@ -0,0 +1,24 @@
...
+
+export class ItemService {
+  private collectionService = new CollectionService()
+
+  public async isOwnedOrManagedBy(
+    id: string,
+    ethAddress: string
+  ): Promise<boolean> {
+    const dbItem = await Item.findOne<ItemAttributes>(id)
+    if (dbItem?.urn_suffix && dbItem?.collection_id) {
```
nicosantangelo, line 13: Nitpick: We could avoid `?.` if you want:

```ts
if (!dbItem) {
  return false
} else if (dbItem.urn_suffix && dbItem.collection_id) {
  return this.collectionService.isOwnedOrManagedBy(
    dbItem.collection_id,
    ethAddress
  )
} else {
  return dbItem.eth_address === dbItem.eth_address
}
```

not sure if more legible, just throwing it out there

nicosantangelo (commented):
🔥

spec/utils.ts
```diff
@@ -72,6 +47,9 @@ export class GenericModel extends Model<any> {}
 // Takes in a mocked (jest.mock()) Model class
 // These methods are sadly order bound, meaning that you'll have to check the router and mock the middlewares in the same order they appear there
 export function mockExistsMiddleware(Table: typeof GenericModel, id: string) {
+  if (!(Table.count as jest.Mock).mock) {
+    throw new Error('Table.count is not mocked')
```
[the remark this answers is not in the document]
LautaroPetaccio: Done!

src/Collection/Collection.service.ts
LautaroPetaccio: It seems to be a good idea! 
I'll do it in another PR so we don't have to do a big conflict resolution do to the multiple merges.

src/Collection/Collection.service.ts
LautaroPetaccio: Done. I've added a method in the `CollectionService` class to replace them.

src/Collection/Collection.service.ts
LautaroPetaccio: I've just removed the getDBCollection method in the utils file (replacing it in the function with its content) to avoid any confusion. Any operations with the collections should come from the Collections service. The utils file has a couple of things that might be part of the CollectionService or another service IMO.

src/Item/Item.service.ts
LautaroPetaccio: It seems pretty much the same to me, but I changed it anyways.

nicosantangelo (approved):
💯

PR #339, opened 2021-10-29
chore: Add checks to upserts of TPW collections

### LINKED ISSUE #317: Add support for updating Third Party Collections
Third party collections can be updated, that is, their URN suffix can be updated only if they don't contain any item that was published or that lives in the TPR under the collection's URN suffix.

LautaroPetaccio:
This PR does the following:
- Adds a missing check for upserting a collection into a third party (that is, a collection that didn't have `urn_suffix` or `third_party_id` now has one).
- Updates the `tpwCollectionURNRegex` to capture the `urn_suffix` that was not being captured before.
- Removes duplicated validation schema that was being already checked using the middleware.
- Adds a check to prevent third party collections with published items from changing the collection's URN.

The PR depends on #327, once it gets merged, the branch will be re-pointed to master.
Closes #317


nicosantangelo (approved):
LGTM

PR #342, opened 2021-10-29
feat: Check for duplicated collection urns

LautaroPetaccio:
This PR checks that the inserted or updates third party collection does not collide with an existing third party collection in the blockchain.

The PR depends on #339 , once it gets merged, the branch will be re-pointed to master.
Closes #1638


src/Collection/Collection.service.ts
```diff
@@ -142,31 +166,35 @@ export class CollectionService {
...
+          collection.third_party_id,
+          collection.urn_suffix
+        )
+
+        // Check if the new URN for the collection already exists
+        await this.checkIfThirdPartyCollectionHasPublishedItems(
+          id,
+          third_party_id,
+          urn_suffix
+        )
```
nicosantangelo, line 185: Would it make sense to do this in a Promise.all?

nicosantangelo (approved):
LGTM, tiny comment

src/Collection/Collection.service.ts
LautaroPetaccio: That is actually a great question. In theory it shouldn't make a request to the graph with the new id's if the older ones already have items published, so it's a sequential procedure, but we could make it parallelized by requesting that data to the graph anyways. What do you think?

PR #344, opened 2021-11-01
feat: Add tiers endpoint

### LINKED ISSUE #338: Create a /item/tiers endpoint to retrieve the third party tiers available
To complete this task, create a `/item/tiers` endpoint that retrieves a list of third party tiers with its [value and price](https://github.com/decentraland/wearables-contracts/pull/37/files#diff-e7460fef9643cafe177f8f297947f06f423c6336d35efd637bbb9b65e7efedddR11).
The endpoint should be accessible without authentication.

LautaroPetaccio:
This PR adds the `/v1/tiers` endpoint that retrieves all tiers.
The PR depends on #342, once it gets merged, the PR will point to master and it can be merged.
Closes #338


src/ethereum/api/thirdParty.ts
```diff
@@ -71,6 +82,18 @@ export class ThirdPartyAPI extends BaseGraphAPI {
     })
     return thirdParties.length > 0
   }
+
+  fetchTiers = async (): Promise<TierFragment[]> => {
+    const {
+      data: { tiers = [] },
+    } = await this.query<{
```
nicosantangelo, line 89: Should this be paginated? It's "free" using `this.paginate()`
I'm not sure how many tiers are there going to be (the `thegraph` max is 1k)

nicosantangelo (commented):
LGTM but left a question (that's why it's not approved yet).

nicosantangelo:
I think the naming is ok, but just in case wanted to bring this up:
Do you think we should be more specific with `Tier`s? I mean, it could be named Item/Collection Tiers or ThirdPartyTiers.

I ask because some other similar concept might come up and we'd have to separate the two. But to be honest, I think tiers is ok for now

src/ethereum/api/thirdParty.ts
LautaroPetaccio: Done. I've paginated the query and changed the path to `/v1/tiers/thirdParty`

nicosantangelo (approved):

PR #346, opened 2021-11-02
chore: Extract and refactor collection service exceptions

LautaroPetaccio:
This PR extracts the collection's service exceptions into a new file.
Instead of receiving an action in a string format, the exceptions now receive an enum.


nicosantangelo (approved):
Awesomeee

PR #347, opened 2021-11-02
fix: allow null urns for DCL collections

nicosantangelo:
@LautaroPetaccio what do you think of this? creating a valid URN for a collection on create (from the front-end) is abit of a hassle and it's not really necessary

As it stands and without this change, creating a DCL collection from the frontend will fail

### DIFF
--- a/src/Collection/Collection.router.ts
+++ b/src/Collection/Collection.router.ts
@@ -29,7 +29,7 @@ import {
 import { CollectionAttributes, FullCollection } from './Collection.types'
 import { upsertCollectionSchema, saveTOSSchema } from './Collection.schema'
 import { hasAccess } from './access'
-import { toFullCollection, isTPCollectionURN } from './utils'
+import { toFullCollection, hasTPCollectionURN } from './utils'
 import { OwnableModel } from '../Ownable/Ownable.types'
 
 const validator = getValidator()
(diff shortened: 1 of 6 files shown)


nicosantangelo:
@LautaroPetaccio I don't particularly like the last commit, but I don't see another way to send urns from the front end on DCL collections, as we don't have the contract_address yet so we can't generate valid ones.
We could send an empty string but I think is a bit obscure

src/Collection/Collection.router.ts
```diff
@@ -380,7 +380,7 @@ export class CollectionRouter extends Router {
     let upsertedCollection: CollectionAttributes
 
     try {
-      if (isTPCollectionURN(collectionJSON.urn)) {
+      if (collectionJSON.urn && isTPCollectionURN(collectionJSON.urn)) {
```
LautaroPetaccio, line 383: What do you think about moving the undefined check for `collectionJSON.urn` into `isTPCollectionURN`? It's a condition to have a non-undefined or non-nulled urn to be a TP collection, so the check could live there and be used elsewhere too.

LautaroPetaccio (changes requested):

LautaroPetaccio (approved):

PR #351, opened 2021-11-04
fix: the urn prefix is already present on third party collection ids

nicosantangelo:
From:
`urn:decentraland:ropsten:collections-thirdparty:urn:decentraland:mumbai:collections-thirdparty:thirdparty2:collection-id`

To
`urn:decentraland:mumbai:collections-thirdparty:thirdparty2:collection-id`

### DIFF
--- a/src/Collection/Collection.router.spec.ts
+++ b/src/Collection/Collection.router.spec.ts
@@ -783,7 +783,7 @@ describe('Collection router', () => {
       thirdPartyDbCollection = {
         ...collectionAttributesMock,
         urn_suffix: 'thesuffix',
-        third_party_id: 'some:third-party-id',
+        third_party_id: 'urn:decentraland:some:third-party-id',
       }
       ;(Collection.find as jest.Mock).mockReturnValueOnce([dbCollection])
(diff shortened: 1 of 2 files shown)


src/Collection/Collection.router.spec.ts
```diff
@@ -783,7 +783,7 @@ describe('Collection router', () => {
       thirdPartyDbCollection = {
         ...collectionAttributesMock,
         urn_suffix: 'thesuffix',
-        third_party_id: 'some:third-party-id',
+        third_party_id: 'urn:decentraland:some:third-party-id',
```
LautaroPetaccio, line 786: Can we put an existing example here?
```suggestion
        third_party_id: 'urn:decentraland:mumbai:collections-thirdparty:third-party-id',
```

LautaroPetaccio (changes requested):

src/Collection/Collection.router.spec.ts
nicosantangelo: Sure

LautaroPetaccio (approved):

PR #352, opened 2021-11-04
feat: return owner (eth_address) on TPW Collection creation

nicosantangelo:
@LautaroPetaccio What do you think of this change?

If we don't do this (as we already do on the [Collection.service](https://github.com/decentraland/builder-server/blob/master/src/Collection/Collection.service.ts#L294)) the front-end can't tell whose the owner of the newly created TP collection thus it only shows up on refresh

### DIFF
--- a/src/Collection/Collection.router.spec.ts
+++ b/src/Collection/Collection.router.spec.ts
@@ -340,7 +340,7 @@ describe('Collection router', () => {
           )
         })
 
-        it('should upsert the collection and respond with the upserted collection', () => {
+        it('should upsert the collection and respond with the upserted collection adding the eth_address', () => {
           return server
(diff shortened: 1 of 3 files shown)


LautaroPetaccio:
Can't we use the returned collection that should have a correct URN set to add the collection and show it?

nicosantangelo:
Not really. The flow looks like this:

```
const remoteCollection: Collection = yield call(() => builder.saveCollection(collection, data))
const newCollection = { ...collection, ...remoteCollection }
```

if we don't add this piece of code, remoteCollection has `owner: ''` (which is correct, but not useful). That means that if you create a new third party collection it'll appear in the state but you won't see it in the UI

PR #353, opened 2021-11-09
hotfix: Fix Bridge.mergeItem which remained with an old implementation

fzavalia:
Hotfix to update the mergeItem function from the bridge.
This function was outdated due to some hotfixes previously made, which were not reverted.


fzavalia:
The failing test is a carry over. It is fixed on lower envs and does not affect prod

cazala (approved):

PR #354, opened 2021-11-09
fix: Third party endpoint without manager not returning third parties

LautaroPetaccio:
This PR does the following:
- Fixes the third party registry controller and query method to return all third parties if a manager is not specified.
- Fixes the `ThirdPartyFragment` to reflect that `maxItems` and `totalItems` are strings.


nicosantangelo (approved):
LGTM

nicosantangelo:
when catch

PR #355, opened 2021-11-16
feat: Add support for deleting TPW items

### LINKED ISSUE #318: Add support for deleting Third Party Items
Third party items can be deleted only if they were not published before.
 To complete this task:
- Add the checks to prevent the items from being deleted if they live in the TPR.

LautaroPetaccio:
This PR adds support for deleting TPW items.
Closes #318

### DIFF
--- a/src/Collection/utils.ts
+++ b/src/Collection/utils.ts
@@ -24,6 +24,10 @@ export function getThirdPartyCollectionURN(
   return `${third_party_id}:${urn_suffix}`
 }
 
+export function isTPCollection(collection: CollectionAttributes): boolean {
+  return collection.third_party_id === null || collection.urn_suffix === null
+}
+
 /**
  * Converts a collection retrieved from the DB into a "FullCollection".
  *
--- a/src/Item/Item.exceptions.ts
+++ b/src/Item/Item.exceptions.ts
@@ -1,0 +1,45 @@
+export enum ItemAction {
+  DELETE = 'deleted',
+}
+
+export enum ItemType {
+  DCL,
+  THIRD_PARTY,
+}
+
+export class NonExistentItemException extends Error {
+  constructor(public id: string) {
+    super("The item doesn't exist.")
+  }
+}
+
+export class ThirdPartyItemAlreadyPublishedException extends Error {
+  constructor(public id: string, public urn: string, action: ItemAction) {
+    super(`The third party item is already published. It can't be ${action}.`)
+  }
+}
+
+export class DCLItemAlreadyPublishedException extends Error {
+  constructor(
+    public id: string,
+    public blockchainItemId: string,
+    public contractAddress: string,
+    action: ItemAction
+  ) {
+    super(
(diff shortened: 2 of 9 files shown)


src/Collection/utils.ts
```diff
@@ -24,6 +24,10 @@ export function getThirdPartyCollectionURN(
   return `${third_party_id}:${urn_suffix}`
 }
 
+export function isTPCollection(collection: CollectionAttributes): boolean {
```
nicosantangelo, line 27: We can use this method on `upsertTPWCollection` too!

```ts
if (collection) {
if (
  collection.third_party_id === null ||
  collection.urn_suffix === null
) {
  throw new WrongCollectionException(
    "The collection can't be converted into a third party collection.",
    { id }
  )
}
```

src/Item/Item.exceptions.ts
```diff
@@ -0,0 +1,45 @@
...
+  }
+}
+
+export class CollectionForItemLockedException extends Error {
+  constructor(public id: string, action: ItemAction) {
+    super(`The collection for the item is locked. The item can't be ${action}.`)
+  }
+}
+
+export class InconsistentItemException extends Error {
```
nicosantangelo, line 41: I know we already did this on `Collection.exceptions.ts` but what do you think of renaming all occurrences of `Exception` to `Error` ?

It's not wrong by any means, it's just that the concept in JS is called `Error`

src/Item/utils.ts
```diff
@@ -25,3 +30,27 @@ export function toDBItem(item: FullItem): ItemAttributes {
...
+  const matches = tpwItemURNRegex.exec(urn)
+  if (matches === null || matches.length !== 2) {
+    throw new Error('The given collection URN is not TWP compliant')
+  }
+
+  return { collectionId: matches[0], itemId: matches[1] }
+}
+
+export function isTPItem(item: ItemAttributes): boolean {
+  return item.urn_suffix !== null
```
nicosantangelo, line 55: Can/Should we use this on `isOwnedOrManagedBy` on `ItemService`?
```ts
} else if (dbItem.urn_suffix && dbItem.collection_id) {
  return this.collectionService.isOwnedOrManagedBy(
    dbItem?.collection_id,
    ethAddress
  )
} else {
```

nicosantangelo (commented):
Gr8

src/Collection/utils.ts
LautaroPetaccio: Used the `isTPCollection` function here.

src/Item/Item.exceptions.ts
LautaroPetaccio: Changed Exception to Error only in the items.

src/Item/utils.ts
LautaroPetaccio: We should, yes! Added it here.

nicosantangelo (approved):

PR #362, opened 2021-11-22
chore: Logs peer & graph API response times

LautaroPetaccio:
This PR logs the peer and graph API response times.

### DIFF
--- a/src/ethereum/api/collection.ts
+++ b/src/ethereum/api/collection.ts
@@ -1,5 +1,7 @@
 import gql from 'graphql-tag'
 import { env } from 'decentraland-commons'
+import { createConsoleLogComponent } from '@well-known-components/logger'
+import { ILoggerComponent } from '@well-known-components/interfaces'
 import {
   collectionFragment,
   itemFragment,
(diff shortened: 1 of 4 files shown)


src/ethereum/api/collection.ts
```diff
@@ -132,6 +135,13 @@ const getRaritiesQuery = () => gql`
 export const COLLECTIONS_URL = env.get('COLLECTIONS_GRAPH_URL', '')
 
 export class CollectionAPI extends BaseGraphAPI {
+  logger: ILoggerComponent.ILogger
+
+  constructor(url: string) {
+    super(url)
+    this.logger = createConsoleLogComponent().getLogger('GraphAPI')
```
nicosantangelo, line 142: Maybe it's worth it to call it `CollectionsGraphAPI` ? or are we trying to group all graph api logs?

I say this because we also have third parties :)

nicosantangelo (approved):
Minor comment

src/ethereum/api/collection.ts
LautaroPetaccio: You're right! Renamed.

PR #364, opened 2021-11-23
feat: Only allow one default cache update at the time

LautaroPetaccio:
This PR changes the way we're updating the default asset packs cache with a mechanism that does the following:
- If there's no cache and someone is retrieving the default asset packs, wait for the asset packs to be retrieved before responding them. This is to prevent responding with an error or invalid data.
- If there's cache and someone is retrieving an updated version of the default asset packs, respond with the old asset packs until the new ones are updated. This is to not queue unnecessary responses.
- If there's no cache to retrieve and there's no-one retrieving the default asset packs to cache them, set a flag to signal that there's a cache update in progress and set a promise so other requests can wait for it to be completed if there's no cache.

### DIFF
--- a/src/AssetPack/AssetPack.router.ts
+++ b/src/AssetPack/AssetPack.router.ts
@@ -45,7 +45,9 @@ export class AssetPackRouter extends Router {
   private lastDefaultAssetPacksFetch: number = 0
   private logger: ILoggerComponent.ILogger
   private isUpdatingDefaultAssetPacksCache: boolean = false
-  private defaultAssetPacksCachePromise: Promise<FullAssetPackAttributes[]>
+  private defaultAssetPacksCachePromise:
+    | Promise<FullAssetPackAttributes[]>
+    | undefined
 
   constructor(router: ExpressApp | express.Router, logger: ILoggerComponent) {
     super(router)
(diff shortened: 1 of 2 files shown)


src/AssetPack/AssetPack.router.ts
```diff
@@ -328,30 +332,40 @@ export class AssetPackRouter extends Router {
...
+      Number(DEFAULT_ASSET_PACK_CACHE)
 
-    if (this.defaultAssetPacks.length === 0 || aDayPassed) {
-      const defaultAssetPacks = await AssetPack.findByEthAddressWithAssets(
+    const defaultAssetPacksIsNotCached =
+      !this.defaultAssetPacksCachedResponse ||
+      this.defaultAssetPacks.length === 0
+
+    // This is to wait for the cache to be completed if there's no cache ready.
+    // It will execute only the first time the server is up, as after the first fetch
```
nicosantangelo, line 344:
```suggestion
    // It will execute only the first time the server is up, and after the first fetch
```

nicosantangelo (approved):
LGTM, mini comment

nicosantangelo (approved):
🚀

PR #367, opened 2021-11-26
feat: Move the upsert of a DCL item to the item service

LautaroPetaccio:
This PR does the following:
- Moves the item upsert code, that inserts or updates an item in the DB to the item service.
- Changes the upsert controller code to use the item service and handle errors.
- Removes an unnecessary DB request to get the collection of the item.

### DIFF
--- a/src/Item/Item.service.ts
+++ b/src/Item/Item.service.ts
@@ -111,10 +111,10 @@ export class ItemService {
     }
   }
 
-  private async checkIfItemIsMovedToAnotherCollection(
+  private checkIfItemIsMovedToAnotherCollection(
     itemToUpsert: FullItem,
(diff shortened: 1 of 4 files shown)


src/Item/Item.service.ts
```diff
@@ -106,4 +110,113 @@ export class ItemService {
       await this.deleteDCLItem(dbItem)
     }
   }
+
+  private checkIfItemIsMovedToAnotherCollection(
```
nicosantangelo, line 114: We can maybe simplify this method to
`checkItemMovedToAnotherCollection`
so we can remove `If` from the method name

nicosantangelo (approved):
Nitpicking, great job

PR #369, opened 2021-12-01
feat: Upsert items with the server's date

LautaroPetaccio:
This PR makes the updated_at and created_at properties of items be set server side.

### DIFF
--- a/src/Item/Item.service.ts
+++ b/src/Item/Item.service.ts
@@ -142,6 +142,7 @@ export class ItemService {
       this.checkItemIsMovedToAnotherCollection(item, dbItem)
     } else {
       item.created_at = new Date()
+      item.updated_at = item.created_at
     }
 
     const collectionId = item.collection_id || dbItem?.collection_id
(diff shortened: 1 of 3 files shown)


src/Item/Item.service.ts
```diff
@@ -138,7 +138,10 @@ export class ItemService {
 
     const dbItem = await Item.findOne<ItemAttributes>(item.id)
     if (dbItem) {
+      item.updated_at = new Date()
       this.checkItemIsMovedToAnotherCollection(item, dbItem)
+    } else {
+      item.created_at = new Date()
```
nicosantangelo, line 144: I think updated_at will be set here regardless, by the decentraland-server lib (to avoid having a NULL updated_at timestamp) maybe we can set it here too?

nicosantangelo (commented):
This PR depends on first merging something on the frontend right? to avoid the request to break

nicosantangelo (approved):

PR #370, opened 2021-12-02
feat: Add HEADER redirect to contents

LautaroPetaccio:
This PR adds a 301 redirect to the contents endpoint to get the sizes of the files without downloading them.

### DIFF
--- a/src/S3/S3Router.spec.ts
+++ b/src/S3/S3Router.spec.ts
@@ -18,14 +18,13 @@ describe('S3 router', () => {
       url = `/storage/contents/${fileName}`
     })
 
-    it('should respond with a cached 301 redirecting to the S3 file', () => {
-      return server
+    it('should respond with a cached 301 redirecting to the S3 file', async () => {
+      await server
         .get(buildURL(url))
         .set(createAuthHeaders('get', url))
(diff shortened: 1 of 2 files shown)


src/S3/S3Router.spec.ts
```diff
@@ -0,0 +1,47 @@
...
+    })
+
+    it('should respond with a cached 301 redirecting to the S3 file', () => {
+      return server
+        .get(buildURL(url))
+        .set(createAuthHeaders('head', url))
+        .expect('Location', `${getBucketURL()}/contents/${fileName}`)
+        .expect('Cache-Control', 'public,max-age=31536000,immutable')
+        .expect(301)
+        .then(() => undefined)
```
nicosantangelo, line 44: Are these `then(() => undefined)` really necessary?

nicosantangelo (approved):
Just a comment for both tests

src/S3/S3Router.spec.ts
LautaroPetaccio: A call to then is needed so the request gets executed (it's what they use to perform it after building it), but using await improves it a lot so I changed it for await.

PR #371, opened 2021-12-03
feat: Add pattern check to the name and the description of the items

LautaroPetaccio:
This PR adds a simple pattern to exclude any names and descriptions of items that disallows using `:` in them.

### DIFF
--- a/src/Item/Item.schema.ts
+++ b/src/Item/Item.schema.ts
@@ -8,8 +8,12 @@ export const itemSchema = Object.freeze({
   properties: {
     id: { type: 'string', format: 'uuid' },
     urn: { type: ['string', 'null'] },
-    name: { type: 'string', maxLength: 32 },
-    description: { type: ['string', 'null'], maxLength: 64 },
+    name: { type: 'string', maxLength: 32, pattern: '^[^:]*$' },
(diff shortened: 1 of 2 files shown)


src/Item/Item.schema.ts
```diff
@@ -8,8 +8,12 @@ export const itemSchema = Object.freeze({
   properties: {
     id: { type: 'string', format: 'uuid' },
     urn: { type: ['string', 'null'] },
-    name: { type: 'string', maxLength: 32 },
-    description: { type: ['string', 'null'], maxLength: 64 },
+    name: { type: 'string', maxLength: 32, pattern: '^[^:]*$' },
+    description: {
+      type: ['string', 'null'],
+      maxLength: 64,
+      pattern: '^[^:]*$',
```
nicosantangelo, line 15: shouldn't these have a `+` instead of an `*` ? or at least a minLength key?

LautaroPetaccio: The only check that the builder has is this one, there's sadly no check about the length. We could add one, but we'll be potentially introducing a breaking change.

nicosantangelo (approved):

PR #374, opened 2021-12-16
feat: split curation in items/curations

nicosantangelo:
This is still kind of a WIP. This works correctly and will not break the front end, it lays the groundwork for the last adjustments we need to make. We create the PR now to get feedback early and reduce the size the of the next ones


src/Curation/Curation.service.ts
```diff
@@ -0,0 +1,116 @@
...
+import { CollectionCuration } from './CollectionCuration'
+import { CurationType } from './Curation.types'
+import { ItemCuration } from './ItemCuration'
+
+export class CurationService<
+  T extends typeof CollectionCuration | typeof ItemCuration
+> {
+  public type: CurationType
+
+  // TODO: Type this better (or remove it completely and use+mock new CurationService)
```
LautaroPetaccio, line 15: Isn't it possible to type it using generics? I'm not sure what you meant with this TODO.

src/Curation/Curation.service.ts
```diff
@@ -0,0 +1,116 @@
...
+        break
+      case ItemCuration:
+        this.type = CurationType.ITEM
+        break
+      default:
+        throw new Error(`Invalid curation model ${CurationModel}`)
+    }
+  }
+
+  async getLatest() {
```
LautaroPetaccio, line 40: What's the type `getLatest` returns? We could use generics to say that this method will return either a CollectionCuration or an ItemCuration.

src/Curation/Curation.service.ts
```diff
@@ -0,0 +1,116 @@
...
+      case CurationType.ITEM:
+        return new CurationService(ItemCuration)
+      default:
+        throw new Error(`Invalid curation type ${type}`)
+    }
+  }
+
+  constructor(public CurationModel: T) {
+    switch (CurationModel) {
+      case CollectionCuration:
```
LautaroPetaccio, line 29: This is comparing the class definition reference, right?

src/Curation/Curation.service.ts
```diff
@@ -0,0 +1,116 @@
...
+      case CurationType.COLLECTION: {
+        const collection = await getMergedCollection(id)
+        return collection && hasCollectionAccess(ethAddress, collection)
+      }
+      case CurationType.ITEM: {
+        const item = await getMergedItem(id)
+        const collection = item.collection_id
+          ? await getMergedCollection(item.collection_id)
+          : undefined
+        return hasItemAccess(ethAddress, item, collection)
```
LautaroPetaccio, line 82: Shouldn't we be testing the item access too?

src/Curation/Curation.router.spec.ts
```diff
@@ -27,29 +49,37 @@ describe('when handling a request', () => {
...
     describe('when the caller belongs to the commitee', () => {
-      it('should resolve with the collections provided by Curation.getAllLatestByCollection', async () => {
+      beforeEach(() => {
+        service = mockService(CollectionCuration)
         mockIsComiteeMember.mockResolvedValueOnce(true)
+      })
 
-        const getAllLatestByCollectionSpy = jest
-          .spyOn(Curation, 'getAllLatestByCollection')
+      it('should resolve with the collections provided by Curation.getAllLatestByCollection', async () => {
```
LautaroPetaccio, line 61:
```suggestion
      it('should resolve with the collections provided by Curation.getLatest', async () => {
```

src/Curation/Curation.router.spec.ts
```diff
@@ -27,29 +49,37 @@ describe('when handling a request', () => {
...
     })
 
     describe('when the caller does not belong to the commitee', () => {
-      it('should resolve with the collections provided by Curation.getAllLatestForCollections', async () => {
+      beforeEach(() => {
+        service = mockService(CollectionCuration)
         mockIsComiteeMember.mockResolvedValueOnce(false)
+      })
 
+      it('should resolve with the collections provided by Curation.getAllLatestForCollections', async () => {
```
LautaroPetaccio, line 82:
```suggestion
      it('should resolve with the collections provided by Curation.getLatestByIds', async () => {
```

src/Curation/Curation.router.spec.ts
```diff
@@ -92,230 +122,420 @@ describe('when handling a request', () => {
...
+
+          expect(updateSpy).toHaveBeenCalledWith(
+            {
+              id: 'curationId',
+              status: 'rejected',
+              updated_at: expect.any(String),
+            },
+            { id: 'curationId' }
+          )
+        })
```
LautaroPetaccio, line 318: We could split the test into two its, one for the return data and the other one for the update expectation.

src/Curation/Curation.router.spec.ts
```diff
@@ -92,230 +122,420 @@ describe('when handling a request', () => {
...
           body: {
             curation: {
               status: 'rejected',
             },
           },
         } as any
+      })
 
-        await expect(router.updateCuration(req)).resolves.toStrictEqual({})
+      describe('when updating a collection', () => {
```
LautaroPetaccio, line 293: What if we add the "curation" word here.
```suggestion
      describe('when updating a collection curation', () => {
```
Or, what if we change it to:
```suggestion
      describe('when the curation to be updated belong to a collection', () => {
```
or:
```suggestion
      describe('and the curation to be updated belong to a collection', () => {
```

The same goes for the other tests that use the same text.

LautaroPetaccio (changes requested):

migrations/1639023686468_create-item-curations.ts
```diff
@@ -0,0 +1,34 @@
...
+      created_at: { type: 'TIMESTAMP', notNull: true },
+      updated_at: { type: 'TIMESTAMP', notNull: true },
+    },
+    {
+      ifNotExists: true,
+      constraints: {
+        foreignKeys: {
+          references: Collection.tableName,
+          columns: 'item_id',
+          onDelete: 'CASCADE',
```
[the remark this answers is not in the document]
nicosantangelo: This was incorrect! fixed, the reference should've been `Item.tableName`

src/Curation/Curation.service.ts
nicosantangelo: The problem here is the return type

```ts
static byType(type: CurationType): CurationService<any> {
```

to type it correctly, `CurationService` should be typed by `CurationType`, like `CurationService<CurationType.COLLECTION>`, but that's kinda hard given that the arg the constructor receives is a Model

```ts
export class CurationService<
  T extends typeof CollectionCuration | typeof ItemCuration
> {

  constructor(public CurationModel: T) {
```

I tried to instantiate `CurationService` using a `CurationType` at first but getting the return type of methods like `getLatestByIds` to be correct wasn't easy

src/Curation/Curation.service.ts
nicosantangelo: Well crap, it seems that at some point along the development of the feature I broke the typings. These are not so trivial to type sadly, I'd have to do something like this:

```ts
// Either this:
async getLatest() {
  const columnName = this.getForeingColumnName()
  return this.getModel().query<T['prototype']['attributes']>( ... )
}

// or:
async getLatest(): Promise<T['prototype']['attributes'][]> {
  const columnName = this.getForeingColumnName()
  return this.getModel().query( ... )
}
```
----------------------

My first implementation revolved around the `CurationType` like this:

```ts
export class CurationService<T extends CurationType> {
  public CurationModel: T extends CurationType.COLLECTION
    ? typeof CollectionCuration
    : typeof ItemCuration

  constructor(public type: T) {
    switch (type as CurationType) {
      case CurationType.COLLECTION:
        this.CurationModel = CollectionCuration
        break
      case CurationType.ITEM:
        this.CurationModel = ItemCuration
        break
      default:
        throw new Error(`Invalid curation model ${type}`)
    }
  }
```

which I think looks cleaner, but I can't get Typescript to agree with me
image

Also, typing `getLatest` in this situation'd look something like

```ts
async getLatest() {
  const columnName = this.getForeingColumnName()
  return this.getModel().query<
    typeof this.CurationModel['prototype']['attributes']
  >(...) 
```

(I think)



Do you have a better idea? which one do you like best?

src/Curation/Curation.service.ts
nicosantangelo: Correct

src/Curation/Curation.service.ts
nicosantangelo: You mean the collection? `hasitemAccess` checks for `isCollectionManager(eth_address, collection)` too!

src/Curation/Curation.router.spec.ts
nicosantangelo: I think I'm going to go with the first one for brevity, as I have to change it for items too

LautaroPetaccio (approved):

PR #376, opened 2021-12-29
feat: TPW item upsert support

LautaroPetaccio:
This PR adds TPW item upsert support.


src/Item/Item.types.ts
```diff
@@ -19,6 +19,7 @@ export type ItemAttributes = {
   id: string // uuid
   /**
    * The urn_suffix field holds the item part of the URN in third party items.
+   * TODO: what about this? The urn_suffix will remain null until the user sets it.
```
LautaroPetaccio, line 22: Should we allow third party items to be created with a null urn_suffix?
These are the two alternatives:
- We could consider that any TWP item upsert MUST have an urn where we will get the suffix from. This implies that the form to create the new items must contain a new field to set the URN.
- We could allow any TPW item upsert to have its URN as null, and let the user set it later.

nicosantangelo: We talked about this a bit, but I think I'd go with option 1. Items MUST have an `urn_suffix`. Maybe we could just pre-fill it with the uuid to avoid having to add a new input to the Modal in the front-end

src/Item/utils.ts
```diff
@@ -37,3 +44,24 @@ export function buildTPItemURN(
...
+  itemURN: string
+): {
+  third_party_id: string
+  network: string
+  collection_urn_suffix: string
+  item_urn_suffix: string
+} {
+  const matches = tpwItemURNRegex.exec(itemURN)
+  if (matches === null) {
+    throw new Error('The given item URN is not TWP compliant')
```
nicosantangelo, line 58:
```suggestion
    throw new Error('The given item URN is not TPW compliant')
```
(or just TP)

src/ethereum/api/Bridge.ts
```diff
@@ -44,11 +45,26 @@ export class Bridge {
...
-  static toFullItem(dbItem: ItemAttributes): FullItem {
+  static toFullItem(
+    dbItem: ItemAttributes,
+    dbCollection?: CollectionAttributes
+  ): FullItem {
+    const hasURN =
+      dbItem.urn_suffix &&
+      dbCollection &&
+      dbCollection.urn_suffix &&
+      dbCollection.third_party_id
```
nicosantangelo, line 56: this could be using isTPCollection right?

nicosantangelo (commented):
Great work

src/Item/Item.service.ts
```diff
@@ -280,14 +308,12 @@ export class ItemService {
...
-      else if (
-        dbItem.collection_id === null &&
-        item.collection_id !== null &&
-        item.urn
-      ) {
-        const decodedItemURN = decodeThirdPartyItemURN(item.urn)
+      } else if (isMovingItemIntoACollection) {
+        // if (item.urn === null) {
+        //   throw new InvalidItemURNError()
+        // }
```
nicosantangelo, line 314: Are these left overs? same on line 334

nicosantangelo (changes requested):

nicosantangelo (approved):

PR #377, opened 2021-12-30
fix: Fix TPW collection upsert manager check

LautaroPetaccio:
This PR does the following:
- Makes the `third_party_id` and the `urn_suffix` unique to prevent logical collections to be overlapped.
- Fixes the management check by providing the third party URN instead of the third party collection URN.
- Changes the way we check for the management of the TP collection by having two different checks, one for the new collection to be inserted, which is providing a new third party id and one for the update process of a collection, which takes into consideration the existing third party id that is stored in the existing collection.


nicosantangelo (changes requested):
I might be missing something, but I don't see how this works. We're now providing a third_party_id to this.isTPWManager but the code for that function didn't seem to change

nicosantangelo (approved):

PR #380, opened 2022-01-06
fix: Item's URN suffix must be unique

LautaroPetaccio:
This PR adds a new migration that makes the item's URN suffix unique.


migrations/1641484605043_make-items-urn-suffix-unique.ts
```diff
@@ -0,0 +1,18 @@
+/* eslint-disable @typescript-eslint/naming-convention */
```
nicosantangelo, line 1: is this necessary?

LautaroPetaccio: It's not! It was put in place by the node-pg-migrate template.

nicosantangelo (approved):

PR #381, opened 2022-01-12
fix: DCL item upsert url

LautaroPetaccio:
This PR changes the way that the `urn_suffix` is built in the DBItem by checking first if it's a TP item.


cazala (approved):
🚀

PR #383, opened 2022-01-14
fix: Remove emojis from title

fzavalia:


src/Forum/client.ts
```diff
@@ -29,3 +35,14 @@ export async function createPost(post: ForumPost): Promise<string> {
...
 }
+
+function sanitizeTitle(title: string) {
+  return removeEmojis(title)
+}
+
+export function removeEmojis(text: string) {
+  return text.replace(
+    /([\u2700-\u27BF]|[\uE000-\uF8FF]|\uD83C[\uDC00-\uDFFF]|\uD83D[\uDC00-\uDFFF]|[\u2011-\u26FF]|\uD83E[\uDD10-\uDDFF])/g,
+    ' '
```
LautaroPetaccio, line 46: Should we leave a blank space where there are emojis or just remove them?

LautaroPetaccio (changes requested):

src/Forum/client.ts
fzavalia: Both for me are the same, added the whitespace to signal that something was actually there.

fzavalia: Removed whitespace

LautaroPetaccio (approved):

PR #386, opened 2022-01-25
feat: added skin category

cazala:
Added `skin` category

### DIFF
--- a/src/Item/wearable/types.ts
+++ b/src/Item/wearable/types.ts
@@ -15,6 +15,7 @@ export enum WearableCategory {
   MASK = 'mask',
   TIARA = 'tiara',
   TOP_HEAD = 'top_head',
+  SKIN = 'skin',
 }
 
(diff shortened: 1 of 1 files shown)


src/Item/wearable/types.ts
```diff
@@ -15,6 +15,7 @@ export enum WearableCategory {
   MASK = 'mask',
   TIARA = 'tiara',
   TOP_HEAD = 'top_head',
+  SKIN = 'skin',
```
LautaroPetaccio, line 18: What if we make the changes needed in the dcl schemas to support skins (and get the enum values and keys) and use it here?

cazala: it is already in dcl-schemas, but I didn't want to replace the local typings in builder and builder-server with those because it would generate a lot of changes and i didn't want to conflict with TPW :/

fzavalia (approved):
🚀

PR #400, opened 2022-02-09
feat: Add endpoint to retrieve available slots for a TP collection

### LINKED ISSUE #387: Add an endpoint to compute available slots
The endpoint should receive the third party id and after checking that the authorized address is a manager of the third party it should return the available slots for that third party by asking the graph the available slot count and subtracting the distinct item curation count for the collection items

Related: https://github.com/decentraland/builder/issues/1748

juanmahidalgo:
Closes #387 

## Description

New endpoint: `/v1/thirdParties/urn:decentraland:mumbai:collections-thirdparty:thirdparty2/slots`

## Response

image

### DIFF
--- a/src/Curation/ItemCuration/ItemCuration.model.ts
+++ b/src/Curation/ItemCuration/ItemCuration.model.ts
@@ -1,4 +1,5 @@
 import { Model, raw, SQL } from 'decentraland-server'
+import { Collection } from '../../Collection'
 import { Item } from '../../Item'
 import { CurationType } from '../Curation.types'
 import { ItemCurationAttributes } from './ItemCuration.types'
@@ -13,5 +14,19 @@ export class ItemCuration extends Model<ItemCurationAttributes> {
     SELECT ic.*
       FROM ${raw(this.tableName)} ic
       INNER JOIN ${raw(Item.tableName)} i ON i.id = ic.item_id AND i.collection_id = ${collectionId}`)
+  }
+
+  static async getItemCurationCountByThirdPartyId(thirdPartyId: string) {
+    return this.query(
+      SQL`SELECT COUNT(DISTINCT ${raw(ItemCuration.tableName)}.id) as Count
+        FROM ${raw(ItemCuration.tableName)}
+        JOIN ${raw(Item.tableName)} ON ${raw(Item.tableName)}.id=${raw(
+        ItemCuration.tableName
+      )}.item_id
+        JOIN ${raw(Collection.tableName)} ON ${raw(
+        Collection.tableName
+      )}.id=${raw(Item.tableName)}.collection_id
+        WHERE ${raw(Collection.tableName)}.third_party_id=${thirdPartyId}`
+    )
   }
 }
--- a/src/Curation/ItemCuration/ItemCuration.model.ts
+++ b/src/Curation/ItemCuration/ItemCuration.model.ts
@@ -16,7 +16,9 @@ export class ItemCuration extends Model<ItemCurationAttributes> {
       INNER JOIN ${raw(Item.tableName)} i ON i.id = ic.item_id AND i.collection_id = ${collectionId}`)
   }
 
-  static async getItemCurationCountByThirdPartyId(thirdPartyId: string) {
+  static async getItemCurationCountByThirdPartyId(
+    thirdPartyId: string
+  ): Promise<{ count: number }[]> {
     return this.query(
       SQL`SELECT COUNT(DISTINCT ${raw(ItemCuration.tableName)}.id) as Count
         FROM ${raw(ItemCuration.tableName)}
--- a/src/Curation/ItemCuration/ItemCuration.model.ts
+++ b/src/Curation/ItemCuration/ItemCuration.model.ts
@@ -16,19 +16,16 @@ export class ItemCuration extends Model<ItemCurationAttributes> {
       INNER JOIN ${raw(Item.tableName)} i ON i.id = ic.item_id AND i.collection_id = ${collectionId}`)
   }
 
-  static async getItemCurationCountByThirdPartyId(
-    thirdPartyId: string
-  ): Promise<{ count: number }[]> {
-    return this.query(
-      SQL`SELECT COUNT(DISTINCT ${raw(ItemCuration.tableName)}.id) as Count
-        FROM ${raw(ItemCuration.tableName)}
-        JOIN ${raw(Item.tableName)} ON ${raw(Item.tableName)}.id=${raw(
-        ItemCuration.tableName
-      )}.item_id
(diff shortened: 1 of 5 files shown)


src/ethereum/api/thirdParty.ts
```diff
@@ -171,6 +179,18 @@ export class ThirdPartyAPI extends BaseGraphAPI {
...
+  fetchMaxItemsByThirdParty = async (thirdPartyId: string): Promise<number> => {
+    const {
+      data: { thirdParties },
+    } = await this.query<{
+      thirdParties: { maxItems: string }[]
+    }>({
+      query: getThirdPartyMaxItems(),
+      variables: { thirdPartyId },
+    })
+    return +thirdParties[0].maxItems
```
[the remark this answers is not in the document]
juanmahidalgo: For sure :) Updated!

src/Curation/ItemCuration/ItemCuration.model.ts
```diff
@@ -14,4 +15,20 @@ export class ItemCuration extends Model<ItemCurationAttributes> {
...
+      SQL`SELECT COUNT(DISTINCT ${raw(ItemCuration.tableName)}.id) as Count
+        FROM ${raw(ItemCuration.tableName)}
+        JOIN ${raw(Item.tableName)} ON ${raw(Item.tableName)}.id=${raw(
+        ItemCuration.tableName
+      )}.item_id
+        JOIN ${raw(Collection.tableName)} ON ${raw(
+        Collection.tableName
+      )}.id=${raw(Item.tableName)}.collection_id
+        WHERE ${raw(Collection.tableName)}.third_party_id=${thirdPartyId}`
+    )
```
nicosantangelo, line 32: This works perfectly, but to avoid having to be so verbose (`raw`s fault) what do you think about using aliases for the tables and re-using them on the ON and WHERE conditions whenever possible?

src/ThirdParty/ThirdParty.router.spec.ts
```diff
@@ -93,4 +95,38 @@ describe('ThirdParty router', () => {
...
+      )
+      ;(thirdPartyAPI.isManager as jest.Mock).mockResolvedValueOnce(true)
+      ;(ItemCuration.getItemCurationCountByThirdPartyId as jest.Mock).mockResolvedValueOnce(
+        [{ count: itemsInCuration }]
+      )
+      url = '/thirdParties/aThirdPartyId/slots'
+    })
+
+    afterEach(() => {
+      jest.restoreAllMocks()
```
nicosantangelo, line 116: If this is being run in a top-level describe of the same scope, it shouldn't be necessary here!

src/ThirdParty/ThirdParty.service.ts
```diff
@@ -0,0 +1,15 @@
...
+import { ThirdParty } from './ThirdParty.types'
+
+export class ThirdPartyService {
+  static async getThirdPartyAvailableSlots(
+    thirdPartyId: ThirdParty['id']
+  ): Promise<number> {
+    const maxItems = await thirdPartyAPI.fetchMaxItemsByThirdParty(thirdPartyId)
+    const itemCurationsCount = await ItemCuration.getItemCurationCountByThirdPartyId(
+      thirdPartyId
+    )
```
nicosantangelo, line 12: You could, in theory, `Promise.all` these methods. Not sure how much sense it makes tho!

src/Curation/ItemCuration/ItemCuration.model.ts
```diff
@@ -14,4 +15,20 @@ export class ItemCuration extends Model<ItemCurationAttributes> {
       FROM ${raw(this.tableName)} ic
       INNER JOIN ${raw(Item.tableName)} i ON i.id = ic.item_id AND i.collection_id = ${collectionId}`)
   }
+
+  static async getItemCurationCountByThirdPartyId(
+    thirdPartyId: string
+  ): Promise<{ count: number }[]> {
+    return this.query(
```
nicosantangelo, line 22: This method is great. Two suggestions:
- In other counts in the project we usually just return the number, to avoid having to burden the caller with knowing that only the first result is relevant (like you did on `fetchMaxItemsByThirdParty`). If you think it looks better we could change it.
- You can type `query` too! so you can do `const r = this.query<{ count: number }>`, where r'd be `{ count: number }[]`

nicosantangelo (commented):
awesome

src/ThirdParty/ThirdParty.router.ts
```diff
@@ -28,4 +38,21 @@ export class ThirdPartyRouter extends Router {
     const fragments = await thirdPartyAPI.fetchThirdPartiesByManager(manager)
     return fragments.map(toThirdParty)
   }
+
+  async getThirdPartyAvailableSlots(req: AuthRequest): Promise<number> {
```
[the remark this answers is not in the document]
juanmahidalgo: Yeah, good idea! I updated the code and added the service as a class variable of the router :)

src/Curation/ItemCuration/ItemCuration.model.ts
juanmahidalgo: Oh absolutely, good call @nicosantangelo! Updated to use aliases wherever is possible, looks way better!

src/ThirdParty/ThirdParty.service.ts
juanmahidalgo: YES! They can be calculated in parallel, so using `Promise.all` here is a really good idea. Updated!

src/ThirdParty/ThirdParty.router.spec.ts
```diff
@@ -93,4 +95,32 @@ describe('ThirdParty router', () => {
...
+    const itemsInCuration = 6
+
+    ;(thirdPartyAPI.fetchMaxItemsByThirdParty as jest.Mock).mockResolvedValueOnce(
+      maxSlots
+    )
+    ;(thirdPartyAPI.isManager as jest.Mock).mockResolvedValueOnce(true)
+    ;(ItemCuration.getItemCurationCountByThirdPartyId as jest.Mock).mockResolvedValueOnce(
+      itemsInCuration
+    )
+    url = '/thirdParties/aThirdPartyId/slots'
```
LautaroPetaccio, line 111: The values should be set in a `beforeEach`, if not, they will not be executed before the `it` clause, but when the file loads.

LautaroPetaccio (changes requested):

src/ThirdParty/ThirdParty.router.spec.ts
juanmahidalgo: Done!!

LautaroPetaccio (approved):

nicosantangelo (approved):

PR #402, opened 2022-02-10
feat: new curation iterations

### LINKED ISSUE #385: Support for publishing TP items directly in the catalyst
We already did a pass on the builder server trying to pick out every place we need to change to make this work. You can see it on [this branch](https://github.com/decentraland/builder-server/tree/feat/new-curation-iterations) and [this commit](https://github.com/decentraland/builder-server/commit/e54eb3be437a96ce40de1ae03b6cbb8fda4d7867)

The idea is to support the new way to generate publications and curations using [a merkle tree](https://hackmd.io/RGDOq2fvQ3mzuKd9cXHhgw)

nicosantangelo:
Closes #385


src/Item/Item.service.ts
```diff
@@ -546,17 +537,9 @@ export class ItemService {
...
-            throw new ThirdPartyItemAlreadyPublishedError(
-              dbItem.id,
-              dbItemURN,
-              ItemAction.UPSERT
-            )
-          }
-
           // Check if the new URN is not already in use
-          if (await thirdPartyAPI.itemExists(itemURN)) {
+          const [wearable] = await peerAPI.fetchWearables([itemURN])
```
LautaroPetaccio, line 541: In theory, we could do the same check by checking if the `urn_suffix` exists for any item on the given collection.

spec/utils.ts
```diff
@@ -165,23 +168,47 @@ export function mockIsThirdPartyManager(
...
   }
 
-  ;(thirdPartyAPI.itemExists as jest.MockedFunction<
-    typeof thirdPartyAPI.itemExists
-  >).mockImplementationOnce((URNToCheck) =>
-    Promise.resolve(URNToCheck === urn && exists)
+  ;(peerAPI.fetchWearables as jest.MockedFunction<
+    typeof peerAPI.fetchWearables
+  >).mockImplementationOnce(([urnToCheck]) =>
+    Promise.resolve(urnToCheck === urn && exists ? [tpWearableMock] : [])
```
LautaroPetaccio, line 211: What if, as we're going to return a mock, we do it by returning something received as a parameter? I mean, we often change the items and such in the tests and this could be a bit of help.

LautaroPetaccio (changes requested):

src/Curation/ItemCuration/ItemCuration.model.ts
```diff
@@ -1,17 +1,53 @@
...
 import { Item } from '../../Item'
-import { CurationType } from '../Curation.types'
+import { CurationStatus, CurationType } from '../Curation.types'
 import { ItemCurationAttributes } from './ItemCuration.types'
 
 export class ItemCuration extends Model<ItemCurationAttributes> {
   static tableName = 'item_curations'
   static type = CurationType.ITEM
 
+  static async exists(itemId: string): Promise<boolean> {
```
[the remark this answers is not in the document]
nicosantangelo: True!

src/Item/utils.ts
```diff
@@ -107,13 +107,13 @@ export async function getMergedItem(id: string): Promise<FullItem> {
       dbCollection.urn_suffix,
       dbItem.urn_suffix
     )
-    const remoteItem = await thirdPartyAPI.fetchItem(urn)
+    const [wearable] = await peerAPI.fetchWearables([urn])
```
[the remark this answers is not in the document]
nicosantangelo: I might've missed this in one of the many conflicts of the rebase 😭 . Changing

src/ethereum/api/Bridge.ts
```diff
@@ -45,24 +49,24 @@ export class Bridge {
...
+    dbItems: ItemAttributes[]
   ): Promise<FullItem[]> {
     const dbTPItemIds = dbItems.map((item) => item.collection_id!)
     const dbTPCollections = await Collection.findByIds(dbTPItemIds)
 
-    const dbTPCollectionsIndex = this.indexById(dbTPCollections)
+    const itemsByURN: Record<string, { item: ItemAttributes }> = {}
+
+    for (const item of dbItems) {
+      const collection = dbTPCollections.find(
```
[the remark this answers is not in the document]
nicosantangelo: 100% bad rebasing skills, no real reasoning. Changing it back. Good find!

src/Item/Item.service.ts
nicosantangelo: That's true. I wasn't 100% sure it was reliable. The database *should* contain all `urn_suffix`es but the ultimate source of truth is the Catalyst.

You think it'd be safe to change it?

spec/utils.ts
LautaroPetaccio: I'm fine with that!

src/Item/Item.service.ts
LautaroPetaccio: It is a good question really, I'm not 100% sure about it.

LautaroPetaccio (approved):

PR #404, opened 2022-02-11
feat: Compute hash on item upsert

### LINKED ISSUE #394: Store the item content hash in the database on upsert
The server currently obtains the content_hash from the blockchain when each item is returned, by using the `Bridge`. To accomodate for #392 we'll need to store the `content_hash` it at root level on each item in the database and check if we can remove the code that merges it with the external sources if necessary

LautaroPetaccio:
This PR adds the `local_content_hash` property to the items table and computes the `local_content_hash` whenever the item is inserted or updated.

Closes #394


src/Item/Item.router.spec.ts
```diff
@@ -1230,6 +1315,49 @@ describe('Item router', () => {
...
+                ...Bridge.toFullItem(dbItem),
+                local_content_hash:
+                  'bafkreiejcahiwp257urwn5u2wn5os3mcjhvid6eqhmok7tezzhvdm5njqy',
+                eth_address: wallet.address,
+                created_at: dbItem.created_at.toISOString(),
+                updated_at: currentDate.toISOString(),
+              },
+              ok: true,
+            })
+          })
```
nicosantangelo, line 1359: Tests look good. Just a question, would it make sense to just use `expect.any(Date)` and avoid having to use fakeTimers?

I know it's less specific, but I also think that the simplicity might be worth it, relegating fakeTimers for real time sensitive tests

src/Item/hashes.ts
```diff
@@ -0,0 +1,101 @@
...
+    metrics: item.metrics,
+  }
+
+  if (!item.rarity) {
+    delete itemMetadata.rarity
+  }
+
+  if (isTPEntity) {
+    // How is the metadata for a TP item built?
+    return omit(itemMetadata, ['collectionAddress'])
```
nicosantangelo, line 66: I think it might be a bit less weird to add the `contractAddress` if `!isTPEntity`?

It's the same in this case, but if we had more code between lines 58 and 64, the type would be "wrong" in between right?
I mean, we're doing `collection.contract_address!` but we know that `contract_address` does not exist in one scenario

nicosantangelo (commented):
⛳

src/Item/hashes.ts
```diff
@@ -0,0 +1,101 @@
...
+    thumbnail: THUMBNAIL_PATH,
+    metrics: item.metrics,
+  }
+
+  if (!item.rarity) {
+    delete itemMetadata.rarity
+  }
+
+  if (isTPEntity) {
+    // How is the metadata for a TP item built?
```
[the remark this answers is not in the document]
LautaroPetaccio: Removed this.

src/Item/hashes.ts
LautaroPetaccio: Made two functions to build the metadata of the entities, one for the TP wearables and the other for the standard wearables.

src/Item/Item.router.spec.ts
```diff
@@ -529,6 +533,13 @@ describe('Item router', () => {
               tpCollectionMock.urn_suffix,
               itemUrnSuffix
             )
+            const originalItemMock = { ...new Item() }
+            mockItem.mockImplementation((createdItem) => {
+              return {
+                ...originalItemMock,
+                upsert: jest.fn().mockResolvedValueOnce(createdItem),
+              } as any
+            })
```
nicosantangelo, line 542: is it necessary to do this? I mean couldn't you just do `Item.update.mockImplementation`? It's reseted afterEach

LautaroPetaccio: We need to test two things:
- The DB updates or inserts the item with the correct information.
- We're returning the upserted item.
To check that we're upserting the item correctly, the only way to do it is by mocking the constructor.

There's another alternative though, and that's to mock the return value of the `upsert` method and then do an assertion on the constructor of the item.

src/Item/Item.types.ts
```diff
@@ -56,3 +57,21 @@ export type FullItem = Omit<ItemAttributes, 'urn_suffix'> & {
   total_supply: number
   content_hash: string | null
 }
+
+type BaseWearableEntityMetadata = {
```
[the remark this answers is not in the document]
LautaroPetaccio: To use `Wearable` we would need to:
- Remove `createdAt` and `updatedAt` from the type for both Standard and TP items
- Remove `collectionAddress` and `rarity` from the TP items.

Is it better to do that or to go with the new types? WDYT?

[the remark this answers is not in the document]
nicosantangelo: Hmm, I guess it depends on how similar we think the types are. By this I mean, if a new prop is added or a property changes from `Date` to `String` to `Wearable` for example, should it be present here?

if yes, maybe its worth it to extend it?

nicosantangelo (approved):

PR #407, opened 2022-02-15
feat: Add DISTINCT ON itemCurations query

### LINKED ISSUE #406: Return only the last ItemCuration for each item
Each endpoint that returns ItemCurations (for a collection or an address) should only return the last ItemCuration for each item, as only one can be active at a time. This could probably be achieved by using DISTINCT ON on SQL

juanmahidalgo:
Closes #406 

## Screenshots
Without the new `DISTINCT ON`, it returns multiple `itemCurations` for the same `itemId`:

image

With the `DISTINCT ON`:
image

It returns the "newest" one.


nicosantangelo (approved):
LGTM

PR #409, opened 2022-02-16
feat: tp publish

### LINKED ISSUE #389: Make publish endpoint TP enabled
The endpoint `/collection/:id/publish` will receive either a TP or a standard collection ID and:

- For a standard collection
  - continue working as is, setting the blockchain item ids of the items
- For a TP collection:
  - It will also receive a signed message as part of the body and store it in the `SlotUsageCheque`
  - It should receive the list of items to be operated on so it can decide where it should create or update the curations for them accordingly (if it's updating the item or it's publishing it).
  - It will publish a forum post designed for the TP collections.

Related: https://github.com/decentraland/builder-server/issues/390

nicosantangelo:
Closes #389


src/Collection/Collection.router.ts
```diff
@@ -244,67 +247,63 @@ export class CollectionRouter extends Router {
   ): Promise<{ collection: FullCollection; items: FullItem[] }> => {
```
nicosantangelo, line 247: This method did not have a `withCollectionAuthorization` but I'm not sure why or if it's necessary

src/Collection/Collection.service.ts
```diff
@@ -23,18 +32,210 @@ import {
...
+    return {
+      collection: Bridge.mergeCollection(dbCollection!, remoteCollection),
+      items: await Bridge.consolidateItems(items, remoteItems),
+    }
+  }
+
+  /**
+   * It should receive the list of items to be operated on so it can decide where it should create or update the curations for them accordingly
+   * It will also receive a signed message and its signature, so it can check and store it in the SlotUsageCheque
+   */
```
LautaroPetaccio, line 135: Would you mind formatting this comment as parameters of the JSDoc?

src/Collection/Collection.service.ts
```diff
@@ -23,18 +32,210 @@ import {
...
+    }
+
+    const allTheSameCollection = dbItems.every(
+      (item) => item.collection_id === collectionId
+    )
+    if (!allTheSameCollection) {
+      throw new InvalidRequestError(
+        'Cannot publish items that belong to different collections'
+      )
+    }
```
LautaroPetaccio, line 180: What if we just compare that `dbCollection.id === item.collection_id` for each item? We could throw a single error too.

src/Collection/Collection.service.ts
```diff
@@ -23,18 +32,210 @@ import {
...
+      promises.push(
+        ItemCuration.create<ItemCurationAttributes>({
+          id: uuid(),
+          item_id: item.id,
+          status: CurationStatus.PENDING,
+          created_at: now,
+          updated_at: now,
+        })
+      )
+    }
```
LautaroPetaccio, line 215: This is a suggestion, but I think that the `map` is perfect for this job as it eliminates the need of doing pushes.
```suggestion
    const promises = dbItems.map(item => 
        ItemCuration.create<ItemCurationAttributes>({
          id: uuid(),
          item_id: item.id,
          status: CurationStatus.PENDING,
          created_at: now,
          updated_at: now,
        })
    )
```

LautaroPetaccio: To create the curations, we could move the `insertCuration` method that we've got in the `Curation.router` into a `Curation.service` and use that.

src/Collection/Collection.service.ts
```diff
@@ -23,18 +32,210 @@ import {
...
+    const itemCurations = await Promise.all(promises)
+    const [lastItemCuration] = itemCurations.slice(-1)
+
+    const collectionCuration = await CollectionCuration.findOne(collectionId)
+    if (collectionCuration) {
+      await CollectionCuration.update(
+        { id: collectionCuration.id },
+        { updated_at: now }
+      )
```
LautaroPetaccio, line 224: If I'm not wrong, we're not updating the `CollectionCuration` whenever we update an ItemCuration in the `Curations.router`, should we move this logic there as well?

LautaroPetaccio (changes requested):

src/Collection/Collection.router.ts
```diff
@@ -244,67 +247,63 @@ export class CollectionRouter extends Router {
...
 
-      const updates = []
+        // Eventually, posting to the forum will be done from the server for both collection types (https://github.com/decentraland/builder/issues/1754)
+        // We should also consider deleteing Forum.router.ts
+        // DCL Collections posts are being handled by the front-end at the moment and the backend updated using '/collections/:id/post'
+        // TODO: Should this be halting the response? Retries?
+        // TODO: Consider slicing the items list if it might get too big
+        const forum_link = await createPost(
+          buildCollectionForumPost(result.collection, result.items)
+        )
```
LautaroPetaccio, line 272: I think that the UI will keep the same retry mechanism for now, so we should be able to dismiss the publishing of the collection if there's no post created so if the post creation fails, the UI can retry it indefinitely.

LautaroPetaccio (changes requested):

src/Item/Item.service.ts
```diff
@@ -121,12 +121,10 @@ export class ItemService {
...
-    const dbCollection = await this.collectionService.getDBCollection(
-      collectionId
-    )
-    const dbItems = await Item.find<ItemAttributes>({
-      collection_id: collectionId,
-    })
+    const [dbCollection, dbItems] = await Promise.all([
+      this.collectionService.getDBCollection(collectionId),
+      Item.find<ItemAttributes>({ collection_id: collectionId }),
+    ])
```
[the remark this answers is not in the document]
nicosantangelo: Couldn't come up with a reason why there where separated and I thought it was an oversight! I think it might be best to split again

src/Forum/utils.ts
```diff
@@ -0,0 +1,70 @@
...
+    title: `Collection '${collection.name}' created by ${shorten(
+      collection.eth_address
+    )} is ready for review!`,
+    raw: `# ${collection.name}
+
+	[View entire collection](${getItemEditorUrl({ collectionId: collection.id })})
+
+	## Wearables
+
+	${items.map(toRawItem).join('\n\n')}`,
```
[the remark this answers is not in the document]
nicosantangelo: yeah you're right. I thought that it wouldn't be necessary at first, but >1k items is too many still. I'll limit it

src/Forum/utils.ts
```diff
@@ -0,0 +1,70 @@
...
+}
+
+function toRawItem(item: FullItem) {
+  const sections = []
+  if (item.description) {
+    sections.push(`- Description: ${item.description}`)
+  }
+  if (item.rarity) {
+    sections.push(`- Rarity: ${item.rarity}`)
+  }
```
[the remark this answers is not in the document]
nicosantangelo: because I'm receiving a FullItem, wouldn't this line work for both cases? I mean if I supply a TP item, rarity will be `undefined` and thus not added, but a Standard item'd work if we use it in the future

src/Forum/utils.ts
```diff
@@ -0,0 +1,70 @@
...
+const BUILDER_SERVER_URL = env.get('BUILDER_SERVER_URL', '')
+
+export function buildCollectionForumPost(
+  collection: CollectionAttributes,
+  items: FullItem[]
+): ForumPost {
+  // We only post in English
+  return {
+    title: `Collection '${collection.name}' created by ${shorten(
+      collection.eth_address
```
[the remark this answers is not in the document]
nicosantangelo: This slipped thru. I guess the third party name is a good option as we don't have the actual collection creator. It hurts a bit to have to add a conditional argument for TPs :/

src/Collection/Collection.service.ts
nicosantangelo: Heh, [there was a comment above this line](https://github.com/decentraland/builder-server/pull/409/commits/9fffc33ff4d5f777ea306d725526ba8a642f9673#diff-0d2b44325541c42451d1866d04950fb50ffcc55ba4c50ec4db19ad16bdd7ef44L206) saying that exact same thing. I figured that it wasn't really necessary as `create` is typed and `insertCuration` had code like this:

```ts
    const curationService = CurationService.byType(type)

    await this.validateAccessToCuration(curationService, ethAddress, id)

    const curation = await curationService.getLatestById(id)
```

which is not really relevant here I think

src/Collection/Collection.service.ts
nicosantangelo: Very possible, I'll check and add it or at least an issue

src/Collection/Collection.router.ts
nicosantangelo: gotcha, I'll remove the `await` then

src/Forum/utils.ts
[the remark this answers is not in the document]
LautaroPetaccio: You're actually right.

src/Forum/utils.ts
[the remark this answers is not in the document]
LautaroPetaccio: I would write `Third Party collection ${collection.name} with URN: ${urn}`
The address of the creator is there to avoid issues with collections that seemed to have the same name due to whitespaces being removed.

src/Collection/Collection.router.ts
```diff
@@ -244,67 +247,63 @@ export class CollectionRouter extends Router {
...
-
-    const isMissingBlockchainItemIds = dbItems.some(
-      (item) => item.blockchain_item_id == null
-    )
-
-    if (isMissingBlockchainItemIds) {
-      remoteItems = await collectionAPI.fetchItemsByContractAddress(
-        dbCollection!.contract_address!
-      )
+    try {
```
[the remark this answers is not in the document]
LautaroPetaccio: You're right, comment dismissed.

src/Curation/Curation.router.spec.ts
```diff
@@ -680,56 +675,15 @@ describe('when handling a request', () => {
...
-              .mockResolvedValueOnce({} as any)
-
-            jest
-              .spyOn(CollectionCuration, 'findByItemId')
-              .mockResolvedValueOnce(undefined)
-          })
-
-          it('should create the virtual collection', async () => {
-            await router.insertItemCuration(req)
+        it('should resolve with the inserted curation', async () => {
```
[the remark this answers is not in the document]
nicosantangelo: I think these are tested elsewhere in this file? I didn't really removed any just added a test to the file!

nicosantangelo:
Updated, but could've missed something 😅. Sorry

src/Collection/Collection.service.ts
```diff
@@ -22,18 +32,205 @@ import {
...
+
+    return {
+      collection: Bridge.mergeCollection(dbCollection!, remoteCollection),
+      items: await Bridge.consolidateItems(items, remoteItems),
+    }
+  }
+
+  /**
+   * It should receive the list of items to be operated on so it can decide where it should create or update the curations for them accordingly
+   * It will also receive a signed message and its signature, so it can check and store it in the SlotUsageCheque
```
LautaroPetaccio, line 134: Would you mind rephrasing it so it explains what the method does?
```suggestion
   * Publishes a TP collection by storing the slots cheque and creating the curations for the items to be published.
```

src/Collection/Collection.service.ts
```diff
@@ -22,18 +32,205 @@ import {
...
+  ): Promise<{
+    collection: CollectionAttributes
+    items: FullItem[]
+  }> {
+    // For DCL collections, once a published collection item changes, the PUSH CHANGES button appears
+    // That will fire a /collections/${collectionId}/curation which will create a new CollectionCuration
+    // Subsequent changes will not show the push changes button, as it's already under_review
+
+    // For TP items, creations always exist. PUSH CHANGES should appear if the item has an approved ItemCuration and has changes in the Catalyst
+    // That should fire /items/:id/curation for each item that changed
```
LautaroPetaccio, line 154: Would you mind moving this into the method's description?

src/Collection/Collection.service.ts
```diff
@@ -22,18 +32,205 @@ import {
...
+
+    // What we should migrate over here is the creation of the virtual CollectionCuration, as there'll always be a publish before a PUSH CHANGES
```
LautaroPetaccio, line 156: Is there a ticket for this? Can we assign it here?

src/Collection/Collection.service.ts
LautaroPetaccio: The access check is not relevant, but checking if there's another curation that's not pending might be (at least in other situation). Saying so, we can see if it needs a refactor later.

LautaroPetaccio (changes requested):

src/Curation/Curation.router.ts
```diff
@@ -229,6 +228,16 @@ export class CurationRouter extends Router {
     const itemId = server.extractFromReq(req, 'id')
     const curationJSON: any = server.extractFromReq(req, 'curation')
     const ethAddress = req.auth.ethAddress
+
+    const { rowCount } = await CollectionCuration.updateByItemId(itemId)
```
LautaroPetaccio, line 232: Review this with @juanmahidalgo .

LautaroPetaccio (approved):

PR #412, opened 2022-02-17
feat: return approval data for collection

### LINKED ISSUE #392: Create the approval data generation endpoint
# New implementation

This endpoint will probably change to let the frontend generate the tree. Will update accordingly with:

The endpoint will be in charge of returning the data that the UI is going to use to calculate the merkle tree for the batch of items to be curated

The endpoint `/collections/:id/approvalData` will  receive the collection id (DB one) and will respond with an array of items which will contain:

```ts
{
  id: string
  content_hash: string
}
```

for all items pending curation for that particular curation.

There's a ton of work already done on https://github.com/decentraland/builder-server/tree/feat/tree-poc done by @nicosantangelo. The branch is currently converting the casing to camelCase, the actual endpoint should return snake_case

# Old implementation

Because the TP collection can have an arbitrary amount of items and the items' list is going to be paginated, the service needs to be in charge of generating the data that the UI is going to use to perform the approval transaction.

The endpoint `/collections/:id/approvalData` will  receive the collection id (DB one) and will respond with:
- The signed message (or cheque) to debit the slots, taken from the `slotUsageCheque` table.
- The merkel root, obtained from the merkle tree where each leaf is generated by:
  - Taking all the hashes of the contents of the item
  - Taking the hash of the item's metadata
  - Ordering all hashes alphanumerically in decreasing order and hashing them all together using Keccak

When building the merkle tree generator, we must take into consideration that the operation might take an arbitrary amount of time that depends on the amount of items. The operation should be done inside a worker.
Check:
- https://nodejs.org/api/worker_threads.html#new-workerfilename-options
- https://github.com/piscinajs/piscina

nicosantangelo:
Adds `/collections/:id/approvalData`

Closes #392

### DIFF
--- a/src/Collection/Collection.router.spec.ts
+++ b/src/Collection/Collection.router.spec.ts
@@ -40,7 +40,12 @@ import {
   CollectionCurationAttributes,
 } from '../Curation/CollectionCuration'
 import { MAX_FORUM_ITEMS } from '../Item/utils'
-import { Item, ItemAttributes, ThirdPartyItemAttributes } from '../Item'
+import {
+  Item,
+  ItemAttributes,
+  ThirdPartyItemAttributes,
+  DBItemApprovalData,
+} from '../Item'
 import { buildCollectionForumPost, createPost } from '../Forum'
 import { SlotUsageCheque } from '../SlotUsageCheque'
 import { CurationStatus } from '../Curation'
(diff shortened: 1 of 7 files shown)


src/Item/Item.model.ts
```diff
@@ -13,6 +14,17 @@ export class Item extends Model<ItemAttributes> {
         WHERE id = ANY(${ids})`)
   }
 
+  static findDBApprovalDataByCollectionId(collectionId: string) {
+    return this.query<DBItemApprovalData>(SQL`
+      SELECT id, urn_suffix, local_content_hash
```
LautaroPetaccio, line 19: Just to be sure:
```suggestion
      SELECT item.id, urn_suffix, local_content_hash
```

src/Item/Item.types.ts
```diff
@@ -59,6 +60,12 @@ export type FullItem = Omit<ItemAttributes, 'urn_suffix'> & {
   content_hash: string | null
 }
 
+export type DBItemApprovalData = Pick<
+  ItemAttributes,
+  'id' | 'urn_suffix' | 'local_content_hash'
+>
+export type ItemApprovalData = Pick<FullItem, 'urn' | 'content_hash'>
```
LautaroPetaccio, line 67: This type will make `content_hash` nullable, but it won't be nullable.

src/Collection/Collection.router.spec.ts
```diff
@@ -1726,4 +1731,248 @@ describe('Collection router', () => {
...
+            )
+            ;(Collection.findOne as jest.Mock).mockResolvedValueOnce(
+              dbTPCollection
+            )
+            ;(Item.findDBApprovalDataByCollectionId as jest.Mock).mockResolvedValueOnce(
+              itemApprovalData
+            )
+          })
+
+          it('should return an array with the data for pending curations', () => {
```
LautaroPetaccio, line 1863:
```suggestion
          it('should respond with a 500 saying that the item is missing some properties', () => {
```

LautaroPetaccio (changes requested):

src/Item/Item.model.ts
```diff
@@ -13,6 +14,17 @@ export class Item extends Model<ItemAttributes> {
...
 
+  static findDBApprovalDataByCollectionId(collectionId: string) {
+    return this.query<DBItemApprovalData>(SQL`
+      SELECT items.id, items.urn_suffix, item_curations.local_content_hash
+        FROM ${raw(this.tableName)} items
+        JOIN ${raw(
+          ItemCuration.tableName
+        )} item_curations ON items.id = item_curations.item_id AND
+        WHERE items.collection_id = ${collectionId}
+          AND item_curations.status = ${CurationStatus.PENDING}`)
```
LautaroPetaccio, line 25: I think we need a DB restriction where an item can't have more than one curation in the pending status. Would you mind creating a ticket for this?

LautaroPetaccio (approved):

src/Item/Item.model.ts
nicosantangelo: Done! https://github.com/decentraland/builder-server/issues/432

PR #414, opened 2022-02-18
feat: Update how is_published for TP items is calculated

### LINKED ISSUE #413: Update how is_published is handled for TP Items
The current implementation of the `Bridge.mergeTPItem` method, considers a TP item as published if it's Catalyst item exists. In practice, this means that, calling the method would guarantee the item is published, as it received a `Wearable` (non nulleable catalyst item) as an argument.

We need to change this logic for the method to consider item curations. An item it's published if it exists any `ItemCuration` for it in the database. The catalyst item could be undefined and it'll show if the item is approved or not

So, for example something like this might work:

```ts
static mergeTPItem(dbItem: ItemAttributes, itemCuration: ItemCuration, catalystItem?: Wearable) {
// (...)
      is_published: true,
      is_approved: !!catalystItem,
```

Important!
We need to check the URN is set correctly

juanmahidalgo:
Closes #413

### DIFF
--- a/src/Item/Item.router.ts
+++ b/src/Item/Item.router.ts
@@ -37,6 +37,8 @@ import {
   UnauthorizedToChangeToCollectionError,
   UnauthorizedToUpsertError,
 } from './Item.errors'
+import { ItemCuration } from '../Curation/ItemCuration'
+import { ItemCurationAttributes } from '../Curation/ItemCuration/ItemCuration.types'
 
 export class ItemRouter extends Router {
   // To be removed once we move everything to the item service
@@ -153,17 +155,18 @@ export class ItemRouter extends Router {
     }
 
     // TODO: We need to paginate this. To do it, we'll have to fetch remote items via the paginated dbItemIds
-    const [allItems, remoteItems] = await Promise.all([
+    const [allItems, remoteItems, , itemCurations] = await Promise.all([
       Item.find<ItemAttributes>(),
       collectionAPI.fetchItems(),
-      thirdPartyAPI.fetchItems(),
+      thirdPartyAPI.fetchItems(), //TODO, UNUSED
+      ItemCuration.find<ItemCurationAttributes>(),
     ])
 
     const { items, tpItems } = this.itemService.splitItems(allItems)
 
     const [fullItems, fullTPItems] = await Promise.all([
       Bridge.consolidateItems(items, remoteItems),
-      Bridge.consolidateTPItems(tpItems),
+      Bridge.consolidateTPItems(tpItems, itemCurations),
     ])
 
     // TODO: sorting (we're not breaking pagination)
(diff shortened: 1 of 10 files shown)


src/Item/Item.router.ts
```diff
@@ -37,6 +37,8 @@ import {
   UnauthorizedToChangeToCollectionError,
   UnauthorizedToUpsertError,
 } from './Item.errors'
+import { ItemCuration } from '../Curation/ItemCuration'
+import { ItemCurationAttributes } from '../Curation/ItemCuration/ItemCuration.types'
```
nicosantangelo, line 41: This can be merged into one I think! and also, can we move them a bit higher? (after all `./`s)

src/ethereum/api/Bridge.ts
```diff
@@ -115,10 +123,20 @@ export class Bridge {
    * @param dbItem - Database TP item
    * @param catalystItem - Catalyst item
    */
-  static mergeTPItem(dbItem: ItemAttributes, catalystItem: Wearable): FullItem {
+  static mergeTPItem(
```
nicosantangelo, line 126: We need to update the JSDoc here

src/ethereum/api/Bridge.ts
```diff
@@ -90,16 +93,21 @@ export class Bridge {
 
     for (const urn in itemsByURN) {
       const item = itemsByURN[urn]
+      const itemCuration = itemCurations.find((ic) => ic.item_id === item.id)
```
nicosantangelo, line 96: I don't want to add further complexity here, but maybe we can index the itemCurations too (as we do collections) so we don't have to iterate multiple times. This is what I have in mind

```ts
    const dbTPCollectionsIndex = this.indexById(dbTPCollections)
    const itemCurationsIndex = this.indexBy(itemCurations, 'item_id')
```

```ts
indexById(list) {
  return this.indexById(list, 'id')
}
indexBy(list, prop) {}
```

What do you think? might be overkill

src/Item/Item.router.ts
```diff
@@ -153,17 +155,18 @@ export class ItemRouter extends Router {
     }
 
     // TODO: We need to paginate this. To do it, we'll have to fetch remote items via the paginated dbItemIds
-    const [allItems, remoteItems] = await Promise.all([
+    const [allItems, remoteItems, , itemCurations] = await Promise.all([
       Item.find<ItemAttributes>(),
       collectionAPI.fetchItems(),
-      thirdPartyAPI.fetchItems(),
+      thirdPartyAPI.fetchItems(), //TODO, UNUSED
```
nicosantangelo, line 161: You can safely remove this one! it's a bad rebase

nicosantangelo (commented):
🇲🇽

src/ethereum/api/Bridge.ts
juanmahidalgo: Yeah, that's a good idea! Implemented :)

nicosantangelo (approved):

PR #419, opened 2022-03-02
feat: Replace signedMessage column for signature

juanmahidalgo:


src/Collection/Collection.router.ts
```diff
@@ -241,70 +245,76 @@ export class CollectionRouter extends Router {
...
+      if (isTPCollection(dbCollection)) {
+        const itemIds = server.extractFromReq<string[]>(req, 'itemIds')
+        const dbItems = await Item.findByIds(itemIds)
+        result = await this.service.publishTPCollection(
+          dbCollection,
+          dbItems,
+          server.extractFromReq(req, 'signedMessage'),
+          server.extractFromReq(req, 'signature'),
+          server.extractFromReq(req, 'qty'),
+          server.extractFromReq(req, 'salt')
```
nicosantangelo, line 271: just a question: Maybe this should be a single object? both for the client hitting the API but for this method's signature. I'm not sure about naming it tho. I was thinking of `cheque` but not sure.
Not blocking the merge, could be another PR

nicosantangelo (approved):
Comment but LGTM

src/Collection/Collection.router.ts
juanmahidalgo: I like the idea! Updated!
`cheque` added to https://github.com/decentraland/builder/pull/1820 as well

nicosantangelo (approved):

PR #420, opened 2022-03-07
feat: Return item curation when updating/publishing

juanmahidalgo:
Please notice that the PR is created against `feat/tp-publish` because it includes many of the changes included in that PR.


src/Collection/Collection.router.ts
```diff
@@ -282,6 +291,7 @@ export class CollectionRouter extends Router {
       return {
         collection: toFullCollection(result.collection),
         items: result.items,
+        itemCurations: result.itemCurations,
```
juanmahidalgo, line 294: Adding this here because we need to keep track of the `itemCurations` in the FE as well, so now it's part of the response.

src/Curation/Curation.router.ts
```diff
@@ -341,13 +341,9 @@ export class CurationRouter extends Router {
...
     }
 
-    return curationService.getModel().update(
-      {
-        ...curation,
-        status: curationJSON.status,
-        updated_at: this.getISODate(),
-      },
-      { id: curation.id }
+    return await curationService.updateStatusAndReturnById(
```
juanmahidalgo, line 344: Same as above, we need the updated row, so I created this method to update the `status` and `updated_at` fields and return the affected row

src/Curation/Curation.types.ts
```diff
@@ -14,7 +14,11 @@ export const patchCurationSchema = Object.freeze({
   properties: {
     status: {
       type: 'string',
-      enum: [CurationStatus.APPROVED, CurationStatus.REJECTED],
+      enum: [
+        CurationStatus.PENDING,
```
juanmahidalgo, line 18: Adding `PENDING` so it can be updated to that status.

src/Collection/Collection.router.ts
```diff
@@ -241,70 +246,77 @@ export class CollectionRouter extends Router {
 
   publishCollection = async (
     req: AuthRequest
-  ): Promise<{ collection: FullCollection; items: FullItem[] }> => {
+  ): Promise<{
+    collection: FullCollection
+    items: FullItem[]
+    itemCurations?: ItemCurationAttributes[]
+  }> => {
```
nicosantangelo, line 253: It might make sense to type this with a `type` of its own? cause is being used [here](https://github.com/decentraland/builder-server/blob/815462f6592d138ebd62aa946f355bb18ce430ab/src/Collection/Collection.router.ts#L249), [on the result definition](https://github.com/decentraland/builder-server/blob/815462f6592d138ebd62aa946f355bb18ce430ab/src/Collection/Collection.router.ts#L259), [and on the service](https://github.com/decentraland/builder-server/blob/815462f6592d138ebd62aa946f355bb18ce430ab/src/Collection/Collection.router.ts#L259)

it could be used in `publishDCLCollection` too.

nicosantangelo (commented):
Small comment

src/Collection/Collection.router.ts
juanmahidalgo: Good idea, I added `PublishCollectionResponse` with a generic for the type of collection to return. 
`publishCollection` returns a `FullCollection` but `this.service.publishTPCollection` returns `CollectionAttributes` for the `collections` attribute so I had to add a generic to the type to support both (mainly because they are not compatible)

PR #428, opened 2022-03-09
feat: Add endpoints to fetch NFTs

### LINKED ISSUE #438: Create endpoints to fetch nfts (as replacement of calling opensea api directly)
These will replace the current calls to the opensea api directly as opensea requires an api key that will be injected by the server when calling these endpoints.

fzavalia:
Closes #438 

OS now requires an api key to use its API.
Now, instead of calling OS's api, clients should call `/v1/nfts` to obtain the data they require.


src/NFT/NFT.router.spec.ts
```diff
@@ -0,0 +1,300 @@
...
+jest.mock('./NFT.service')
+
+const mockNFTService = NFTService as jest.MockedClass<typeof NFTService>
+
+const server = supertest(app.getApp())
+
+const mockAddress = '0x6D7227d6F36FC997D53B4646132b3B55D751cc7c'
+
+describe('when getting nfts', () => {
+  describe('when owner query param does not have a valid etherum address pattern', () => {
```
cazala, line 16:
```suggestion
describe('when owner query param does not have a valid ethereum address pattern', () => {
```

src/NFT/NFT.router.spec.ts
```diff
@@ -0,0 +1,300 @@
...
+        first: 10,
+        owner: mockAddress,
+        skip: 10,
+      })
+    })
+  })
+})
+
+describe('when getting a single nft', () => {
+  describe('when contractAddress path param does not have a valid etherum address pattern', () => {
```
cazala, line 186:
```suggestion
  describe('when contractAddress path param does not have a valid ethereum address pattern', () => {
```

cazala (approved):
🏌️‍♂️

PR #429, opened 2022-03-09
fix: TP Manager check

LautaroPetaccio:
This PR fixes the manager check by changing the `isManager` query.
The PR also includes a change to bridge TP and Standard collections asynchronously.


src/ethereum/api/thirdParty.ts
```diff
@@ -79,13 +79,8 @@ const getItemQuery = () => gql`
...
 const isManagerQuery = () => gql`
   query isManager($thirdPartyId: String!, $managers: [String!]) {
-    thirdParties(
-      first: 1
-      where: { thirdParty: $thirdPartyId, searchCollectionId: $collectionId }
-      orderBy: reviewedAt
-      orderDirection: desc
-    ) {
-      ...thirdPartyItemFragment
+    thirdParties(first: 1, where: { id: $thirdPartyId }) {
```
nicosantangelo, line 82: I can see this wasn't working correctly (probably a rebase this file was butchered) but the solution doesn't seem to work either?

the method is receiving both the thirdPartyId and an array of managers to check against, but is just using the id

nicosantangelo (changes requested):

nicosantangelo (approved):

PR #430, opened 2022-03-10
fix: Query for finding non TP items by owner in `/:address/items`

fzavalia:
`Item.find<ItemAttributes>({ eth_address, urn_suffix: null })` 
Was being mapped to 
`SELECT ... WHERE urn_suffix = null` 
Instead of
`SELECT ... WHERE urn_suffix IS NULL` 
So no results would be returned.


juanmahidalgo (approved):
glad you found this @fzavalia 👍

PR #431, opened 2022-03-10
feat: rollback publish changes

### LINKED ISSUE #422: Add a rollback mechanism to TP publishes
If anything goes wrong midway into the publication of a collection, we should revert the operation and any changes made to the database, so the front-end can retry gracefully without any intermediate states

nicosantangelo:
Closes #422

### DIFF
--- a/src/Collection/Collection.router.spec.ts
+++ b/src/Collection/Collection.router.spec.ts
@@ -1374,6 +1374,84 @@ describe('Collection router', () => {
         })
       })
 
+      describe('and interating with the database fails', () => {
+        let slotUsageChequeDeleteSpy: jest.SpyInstance<
+          ReturnType<typeof SlotUsageCheque.delete>
+        >
+        let itemCurationDeleteSpy: jest.SpyInstance<
+          ReturnType<typeof ItemCuration.delete>
+        >
+
+        beforeEach(() => {
+          ;(Item.findByIds as jest.Mock).mockResolvedValueOnce([dbItemMock])
+          ;(ItemCuration.findLastCreatedByCollectionIdAndStatus as jest.Mock).mockResolvedValueOnce(
+            undefined
+          )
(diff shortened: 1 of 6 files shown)


src/Collection/Collection.router.spec.ts
```diff
@@ -1374,6 +1374,78 @@ describe('Collection router', () => {
...
+        >
+
+        beforeEach(() => {
+          ;(Item.findByIds as jest.Mock).mockResolvedValueOnce([dbItemMock])
+          ;(ItemCuration.findLastCreatedByCollectionIdAndStatus as jest.Mock).mockResolvedValueOnce(
+            undefined
+          )
+          jest.spyOn(ethers.utils, 'verifyMessage').mockReturnValue('0x')
+
+          jest.spyOn(SlotUsageCheque, 'create').mockResolvedValueOnce({})
```
LautaroPetaccio, line 1389: Would you mind adding here an id so we can later assert that the delete procedure was called with this id?

src/Collection/Collection.router.spec.ts
```diff
@@ -1374,6 +1374,78 @@ describe('Collection router', () => {
         })
       })
 
+      describe('and interating with the database fails', () => {
+        let itemCurationDeleteSpy: jest.SpyInstance<
+          ReturnType<typeof ItemCuration.delete>
+        >
```
LautaroPetaccio, line 1380: With the change, instead of `ItemCuration.delete`, `ItemCuration.deleteByIds` is being used.

src/Collection/Collection.router.spec.ts
LautaroPetaccio: The same goes for the `delete` method of the `SlotUsageCheque`.

LautaroPetaccio (changes requested):

LautaroPetaccio (approved):

PR #433, opened 2022-03-11
fix: Find third party ids query

LautaroPetaccio:
This PR fixes an issue where we were using `item` instead of `items` when naming a table.
The aliases were removed, as the names of the collections are sufficient enough for this query and the table names were replaced by their table variable.

### DIFF
--- a/migrations/migrate.ts
+++ b/migrations/migrate.ts
@@ -17,6 +17,7 @@ export function migrate(
   }
 
   const spawnArgs = [
+    '--no-check-order',
     '--database-url-var',
     'CONNECTION_STRING',
     '--migration-file-language',
--- a/migrations/migrate.ts
+++ b/migrations/migrate.ts
@@ -17,7 +17,6 @@ export function migrate(
   }
 
   const spawnArgs = [
-    '--no-check-order',
     '--database-url-var',
     'CONNECTION_STRING',
     '--migration-file-language',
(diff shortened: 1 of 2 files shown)


src/Item/Item.model.ts
```diff
@@ -33,13 +33,14 @@ export class Item extends Model<ItemAttributes> {
...
     return this.query<ItemAttributes>(SQL`
-      SELECT items.*
-        FROM ${raw(this.tableName)} items 
-        JOIN ${raw(
-          Collection.tableName
-        )} collections ON collections.id = item.collection_id
-        WHERE collections.third_party_id = ANY(${thirdPartyIds})`)
+      SELECT ${itemsTable}.*
+        FROM ${itemsTable}
+        JOIN ${collectionsTable} ON collections.id = ${itemsTable}.collection_id
```
nicosantangelo, line 42: missing a ${collectionsTable} here right? (`collections.id`)

Question: it's a good idea to do this in general to avoid typos, but at the same time it makes the query really difficult to read. Can we strike a balance perhaps? shorter names but using the string in the query?

(we might need tests for queries heh)

LautaroPetaccio: Ups, added it!
What makes me doubt is that we have the table name used as a variable for what I think is to avoid typos (or be reactive when the table changes), but we're only using it for the table once.

migrations/migrate.ts
```diff
@@ -17,6 +17,7 @@ export function migrate(
   }
 
   const spawnArgs = [
+    '--no-check-order',
```
nicosantangelo, line 20: 👀

nicosantangelo (approved):

PR #436, opened 2022-03-15
feat: simplify approval data endpoint to return only content hashes

nicosantangelo:
Related to https://github.com/decentraland/content-hash-tree/commit/b27c7ff451ccd2ab9913ac8e0bbf257f3d8408a9

Needed for:
- https://github.com/decentraland/builder/issues/1846
- https://github.com/decentraland/builder/issues/1847

It changes the 
`/collections/:id/approvalData` endpoint to only return contentHashes instead of the `{ content_hash; urn }` object

### DIFF
--- a/src/Collection/Collection.router.spec.ts
+++ b/src/Collection/Collection.router.spec.ts
@@ -1891,13 +1891,11 @@ describe('Collection router', () => {
               {
                 id: uuid(),
                 local_content_hash: 'Qm1abababa',
-                urn_suffix: '1',
               },
-              { id: uuid(), local_content_hash: '', urn_suffix: '2' },
+              { id: uuid(), local_content_hash: '' },
               {
                 id: uuid(),
                 local_content_hash: 'Qm3rererer',
-                urn_suffix: '',
               },
             ]
 
@@ -1928,7 +1926,7 @@ describe('Collection router', () => {
                     eth_address: wallet.address,
                   },
                   error:
-                    'Item missing the urn_suffix or local_content_hash needed to approve it',
+                    'Item missing the local_content_hash needed to approve it',
                 })
               })
           })
@@ -1940,17 +1938,14 @@ describe('Collection router', () => {
               {
                 id: uuid(),
                 local_content_hash: 'Qm1abababa',
-                urn_suffix: '1',
               },
               {
                 id: uuid(),
                 local_content_hash: 'Qm2bdbdbdb',
-                urn_suffix: '2',
               },
               {
                 id: uuid(),
                 local_content_hash: 'Qm3rererer',
-                urn_suffix: '3',
               },
             ]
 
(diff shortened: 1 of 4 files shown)


src/Item/Item.types.ts
```diff
@@ -63,9 +63,16 @@ export type FullItem = Omit<ItemAttributes, 'urn_suffix'> & {
 
 export type DBItemApprovalData = Pick<
   ItemAttributes,
-  'id' | 'urn_suffix' | 'local_content_hash'
+  'id' | 'local_content_hash'
 >
-export type ItemApprovalData = Pick<FullItem, 'urn'> & { content_hash: string }
+export type ItemApprovalData = {
+  cheque: {
```
nicosantangelo, line 69: It bothers me that I can't really type this the same as `PublishCheque`

```ts
export type PublishCheque = {
  signedMessage: string
  signature: string
  qty: number
  salt: string
}

// or maybe as SlotUsageChequeAttributes
```

but I can't seem to actually generalize it

src/Item/Item.model.ts
```diff
@@ -16,7 +16,7 @@ export class Item extends Model<ItemAttributes> {
 
   static findDBApprovalDataByCollectionId(collectionId: string) {
     return this.query<DBItemApprovalData>(SQL`
-      SELECT items.id, items.urn_suffix, item_curations.local_content_hash
+      SELECT items.id, item_curations.local_content_hash
```
juanmahidalgo, line 19:
```suggestion
      SELECT items.id, item_curations.content_hash
```
I believe the column is called `content_hash`

src/Item/Item.types.ts
juanmahidalgo: What about using `cheque: Omit<PublishCheque, 'signedMessage'>` ? I'd be OK with that :)

src/SlotUsageCheque/SlotUsageCheque.model.ts
```diff
@@ -1,7 +1,23 @@
...
+
+  static async findLastByCollectionId(
+    collectionId: string
+  ): Promise<SlotUsageChequeAttributes | undefined> {
+    const result = await this.query<SlotUsageChequeAttributes>(SQL`
+      SELECT slot_usage_cheque.*
+      FROM ${raw(this.tableName)} slot_usage_cheque
+      INNER JOIN ${raw(
+        Collection.tableName
+      )} collection ON collection.id = slot_usage_cheque.collection_id AND collection.collection_id = ${collectionId}
```
juanmahidalgo, line 17:
```suggestion
      )} collection ON collection.id = slot_usage_cheque.collection_id AND collection.id = ${collectionId} 
```
I think you want `collection.id` instead of `collection.collection_id`

src/Collection/Collection.router.spec.ts
```diff
@@ -1374,9 +1375,9 @@ describe('Collection router', () => {
       })
 
       describe('when the supplied data is correct', () => {
-        let slotUsageChequeCreateSpy: jest.SpyInstance<
-          ReturnType<typeof SlotUsageCheque.create>
-        >
+        // let slotUsageChequeCreateSpy: jest.SpyInstance<
```
juanmahidalgo, line 1378: 😸

juanmahidalgo (approved):
looks great 🚀

PR #437, opened 2022-03-15
feat: Approval data return a dictionary instead of array of hashes

juanmahidalgo:


src/Item/Item.model.ts
```diff
@@ -20,7 +20,7 @@ export class Item extends Model<ItemAttributes> {
         FROM ${raw(this.tableName)} items
         JOIN ${raw(
           ItemCuration.tableName
-        )} item_curations ON items.id = item_curations.item_id AND
+        )} item_curations ON items.id = item_curations.item_id
```
LautaroPetaccio, line 23: I was thinking that the tree to be built requires not only this data, but the data of all the items that are published for a third party. We need all the data to re-build the tree, and the tree is built with all the items of a third party.

LautaroPetaccio (changes requested):

src/Item/Item.model.ts
nicosantangelo: True, I think then that the query should not check for curation status but rather for curation "existence", which is already being done by the INNER JOIN right?

nicosantangelo (approved):
The changes are 💯 
The conversation [here](https://github.com/decentraland/builder-server/pull/437/files#r827306650) can be done in another issue

src/Item/Item.model.ts
juanmahidalgo: Alright, as Nico suggested, we can take care of that in another PR. Issue created: https://github.com/decentraland/builder-server/issues/439

PR #442, opened 2022-03-16
fix: Simplify curation access

LautaroPetaccio:


src/Collection/Collection.model.ts
```diff
@@ -25,6 +26,18 @@ export class Collection extends Model<CollectionAttributes> {
         WHERE contract_address = ANY(${contractAddresses})`)
   }
 
+  static async findCollectionOwningItem(
```
nicosantangelo, line 29: I like this name, but I think we were using ByXXX if you want to be consistent ( `findByItemId` )

src/Collection/access.ts
```diff
@@ -25,7 +25,7 @@ export async function hasAccess(
     isCommitteeMember(eth_address),
     isTPCollection(collection)
       ? thirdPartyAPI.isManager(collection.third_party_id, eth_address)
-      : isManager(eth_address, collection),
+      : !!collection.is_published && isManager(eth_address, collection),
```
nicosantangelo, line 28: I don't know if I understand this condition:
- Why are we casting a (supposedly) boolean value? might be badly typed
image
- For what I can tell, this is so checking for managers makes sense. Meaning that only published collections have managers. But reading the code, shouldn't `isManager()` just return false then?

src/Curation/Curation.router.ts
```diff
@@ -239,12 +239,30 @@ export class CurationRouter extends Router {
...
+      if (error instanceof NonExistentCollectionError) {
+        throw new HTTPError(
+          'Not found',
+          { id: error.id, ethAddress },
+          STATUS_CODES.notFound
+        )
+      } else if (error instanceof HTTPError) {
+        throw error
+      } else {
+        throw new HTTPError(
```
nicosantangelo, line 259: Question! why is this the only place where we're doing this? meaning checking for HTTPErrors and throwing a new one otherwise.

It seems like the code in insertCuration should do the same ( `if (error instanceof HTTPError)` )? and the other catches might need to throw a new HTTPError instead of `throw error` ?

src/Item/Item.model.ts
```diff
@@ -1,5 +1,5 @@
 import { Model, SQL, raw } from 'decentraland-server'
-import { Collection } from '../Collection'
+import { Collection } from '../Collection/Collection.model'
```
nicosantangelo, line 2: image

circular dep?

src/Item/access.ts
```diff
@@ -37,7 +50,8 @@ export async function hasAccess(
         eth_address
       )
     } else {
-      isManager = isCollectionManager(eth_address, collection)
+      isManager =
+        collection.is_published && isCollectionManager(eth_address, collection)
```
nicosantangelo, line 54: same question as [here](https://github.com/decentraland/builder-server/pull/442/files#diff-4f08a39b94a0e600623e156bc99bf67fc48b8cb01d81aacfbb15cf14ccdf35fbR28)

Also, reading here, it seems odd that checking for `is_published` seems to be required for calling `isManager`, if it's a MUST maybe it should be in the function?

nicosantangelo (commented):
Awesome !

src/Item/Item.model.ts
LautaroPetaccio: Sadly there are a lot of circular dependencies between item and collection. I can't remember if this one was impacting our tests, but I tried to reduce them as much as possible.

src/Item/access.ts
LautaroPetaccio: The problem is that there's an issue with the types here. The method is taking `CollectionAttributes` as the parameter for its type, but `CollectionAttributes` is not correctly typed, it has properties like `managers`, `is_published` and `is_approved` that don't live in the database, making `is_published` possibly null.
We need to fix the collection types to be able to represented a `MergedCollection` containing all of these values and a real `CollectionAttributes` object, but this PR has gotten pretty long already.

src/Collection/access.ts
LautaroPetaccio: The thing is that isManager should only be called with a merged collection. The code should be something like ```isMergedCollection(collection) && isManager(eth_address, collection)```.
I could add this function elsewhere, but I would love to do it with some type changes to collection first.
See related comment down below.

src/Collection/Collection.model.ts
LautaroPetaccio: Changed it to `findByItemId`!

src/Curation/Curation.router.ts
LautaroPetaccio: You're right! moved it to `validateAccessToCuration` so the errors can get catched every time.

PR #443, opened 2022-03-17
feat: Use hashing lib to hash TP & Standard entities

LautaroPetaccio:
This PR does the following:
- Hashes the TP item entity in a different way to accommodate to the ADR62
- Hashes the Standard item entity using the @dcl/hashing lib.


src/Item/hashes.ts
```diff
@@ -11,6 +6,7 @@ import {
 } from './Item.types'
 import { buildTPItemURN, getDecentralandItemURN, isTPItem } from './utils'
 import { isTPCollection } from '../Collection/utils'
+import { calculateMultipleHashesADR32, keccak256Hash } from '@dcl/hashing'
```
nicosantangelo, line 9: 🔝

nicosantangelo (approved):

PR #444, opened 2022-03-17
feat: check signer against publish request sender

### LINKED ISSUE #445: Check the signer against the publish request sender

nicosantangelo:
We where only checking for valid signatures

Closes #445

### DIFF
--- a/src/Collection/Collection.service.ts
+++ b/src/Collection/Collection.service.ts
@@ -154,7 +154,7 @@ export class CollectionService {
   public async publishTPCollection(
     dbCollection: ThirdPartyCollectionAttributes,
     dbItems: ThirdPartyItemAttributes[],
-    signer: string,
+    signerAddress: string,
     cheque: PublishCheque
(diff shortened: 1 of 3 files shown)


src/Collection/Collection.service.ts
```diff
@@ -172,10 +173,16 @@ export class CollectionService {
     }
 
     try {
-      ethers.utils.verifyMessage(signedMessage, signature) // Throws if invalid
+      const address = ethers.utils.verifyMessage(signedMessage, signature) // Throws if invalid
+
+      if (signer.toLowerCase() !== address.toLowerCase()) {
```
nicosantangelo, line 178: Maybe we should actually check the address belongs to the managers list of the collection?

LautaroPetaccio: We can, but the contract is already doing that on its own.

LautaroPetaccio (approved):

PR #446, opened 2022-03-18
fix: Fix find collection query & params order

juanmahidalgo:


src/Curation/Curation.router.ts
```diff
@@ -408,12 +408,12 @@ export class CurationRouter extends Router {
 
   private validateAccessToCuration = async (
     service: CurationService<any>,
-    id: string,
-    ethAddress: string
+    ethAddress: string,
```
juanmahidalgo, line 411: the function was being called with the `ethAddress` as the 2nd parameter. It was working correctly but only because they were being send in the wrong order to the `hasAccess` down below.

LautaroPetaccio: Great catch!

LautaroPetaccio (approved):

PR #448, opened 2022-03-18
chore: Extend NFT type

fzavalia:
In order to cover more cases, a lot more fields have been added to the NFT type.
Based both on what is being consumed by [unity-renderer](https://github.com/decentraland/unity-renderer) with [these types](https://github.com/decentraland/unity-renderer/blob/master/unity-renderer/Assets/Scripts/MainScripts/DCL/ServiceProviders/OpenSea/Types.cs#L12) and what is being consumed by the [builder](https://github.com/decentraland/builder) in the current [asset sagas](https://github.com/decentraland/builder/blob/master/src/modules/asset/sagas.ts#L25).


cazala (approved):
🤸

PR #451, opened 2022-03-18
fix: Use older hash for standard items

LautaroPetaccio:
Change the item entity hashing method to use the older method until we switch to the new one.


nicosantangelo (changes requested):
Rememba the test

cazala (approved):

nicosantangelo (approved):

PR #453, opened 2022-03-18
feat: emotes

cazala:
Adds support for Emote schemas

### DIFF
--- a/src/Item/Item.schema.ts
+++ b/src/Item/Item.schema.ts
@@ -32,7 +32,7 @@ export const itemSchema = Object.freeze({
     created_at: { type: 'string' },
     updated_at: { type: 'string' },
     type: { enum: Object.values(ItemType) },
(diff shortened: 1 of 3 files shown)


src/Item/Item.schema.ts
```diff
@@ -31,7 +32,7 @@ export const itemSchema = Object.freeze({
     created_at: { type: 'string' },
     updated_at: { type: 'string' },
     type: { enum: Object.values(ItemType) },
-    data: { type: 'object', oneOf: [wearableSchema] },
+    data: { type: 'object', anyOf: [wearableSchema, emoteSchema] },
```
cazala, line 35: I changed `oneOf` to `anyOf` otherwise lots of tests fail, because some of the data used on test cases are both valid wearable AND emote schemas, and `oneOf` throws if there's more than 1 schema passing. `anyOf` allows both schemas to pass.

LautaroPetaccio (approved):

PR #454, opened 2022-03-21
fix: Hash computation over older item

LautaroPetaccio:
This PR moves the computation of the `local_content_hash` of the item down below the `upsertDCLItem` method so the `local_content_hash` gets computed with the changes to the item.

### DIFF
--- a/src/Item/Item.service.ts
+++ b/src/Item/Item.service.ts
@@ -360,7 +360,6 @@ export class ItemService {
     dbCollection: CollectionAttributes | undefined,
     eth_address: string
   ): Promise<FullItem> {
-    let contentHash = null
     const canUpsert = await new Ownable(Item).canUpsert(item.id, eth_address)
     if (!canUpsert) {
       throw new UnauthorizedToUpsertError(item.id, eth_address)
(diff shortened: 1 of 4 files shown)


src/Item/Item.service.ts
```diff
@@ -433,7 +429,13 @@ export class ItemService {
       eth_address,
     })
 
-    attributes.local_content_hash = contentHash
+    attributes.blockchain_item_id = dbItem ? dbItem.blockchain_item_id : null
+
+    // Compute the content hash of the item to later store it in the DB
+    attributes.local_content_hash =
+      attributes?.blockchain_item_id && dbCollection?.contract_address
```
nicosantangelo, line 436: two things

- is `attributes?` necessary? seems to always be defined right?
- shouldn't this be a function with a name? it's also used elsewhere like in `buildStandardWearableEntityMetadata`

LautaroPetaccio: I didn't understand the question about attributes. I'm using attributes there to have an updated representation of what's going to be inserted into the database to compute its hash.
About the second one, it's similar to the one in `buildStandardWearableEntityMetadata`, as this one checks that the `dbCollection` exists, but I've refactored into a function so we can name it.

LautaroPetaccio: Oh, you're right! I've changed the code to use the function that checks if the item was published, this code is outdated

nicosantangelo (approved):

PR #458, opened 2022-03-22
Fix: fix how the signature is being validated

### LINKED ISSUE #456: The server can't get the correct address after a tp publish
When publishing a TP collection, we're using the method `.verifyMessage` from ethers with a signedMessage and the signature, but it's not enough to get the correct address. This leads to the collection always failing upon publication.

We should update the method so it returns the correct address

juanmahidalgo:
Closes #456


src/Collection/Collection.router.spec.ts
```diff
@@ -1381,8 +1352,11 @@ describe('Collection router', () => {
           itemIds = items.map((item) => item.id)
           ;(Item.findByIds as jest.Mock).mockResolvedValueOnce(items)
           jest
-            .spyOn(ethers.utils, 'verifyMessage')
-            .mockReturnValue(wallet.address)
+            .spyOn(ethers.utils, 'verifyTypedData')
+            .mockImplementationOnce(() => wallet.address)
```
LautaroPetaccio, line 1356: Didn't it work with `mockReturnValue`?

LautaroPetaccio (approved):
Great job!

src/Collection/Collection.router.spec.ts
juanmahidalgo: Yeah, that one works too. Updated!

PR #460, opened 2022-03-23
feat: Add logic to update forum post if it already exists

### LINKED ISSUE #452: Update or skip the forum post after the first item curation batch
When the first batch of TP Items is published, the frontend will call:
`/collections/:id/publish`

which in turn will publish the forum post for the collection and update the `forum_link` property on the Collection itself.

After this first publish, the collection might have additional batches of items to publish, which will call `/collections/:id/publish`.

So we need to:
- [ ] Stop the endpoint from blowing up after subsequent requests
- [ ] Investigate if we can update it with some relevant information

juanmahidalgo:
Closes #452


nicosantangelo (approved):
LGTM

Should we create an issue to update the forum_id from the front-end on standard collections for the time being?
here: https://github.com/decentraland/builder/blob/master/src/modules/forum/sagas.ts#L24

juanmahidalgo:
> LGTM
> 
> Should we create an issue to update the forum_id from the front-end on standard collections for the time being? here: https://github.com/decentraland/builder/blob/master/src/modules/forum/sagas.ts#L24

We definitely need to create a follow-up ticket for the standard collections. I'd say to create one to deprecate the action on the FE and handle the creation as we do with the TP items, when publishing in the server. What do you think?

PR #461, opened 2022-03-23
fix: Allow TP approval flow without cheque

LautaroPetaccio:
This PR does the following:
- Prevents a new item curation to be created if the item was not curated before, removing the possibility of publishing new TP items from a different endpoint than the publish one.
- Returns in the approval data endpoint a new property signaling if the cheque was used or not to decide if it should be consumed.


src/Collection/Collection.service.ts
```diff
@@ -446,13 +450,23 @@ export class CollectionService {
...
     }, {} as Record<string, string>)
 
+    const slotUsageCheckHash = await getChequeMessageHash(
+      slotUsageCheque,
+      slotUsageCheque.third_party_id
+    )
+
+    const remoteCheque = await thirdPartyAPI.fetchReceiptById(
+      slotUsageCheckHash
+    )
```
nicosantangelo, line 460: The remote cheque not existing is something that could happen? or should we throw here?

nicosantangelo (approved):

src/Collection/Collection.service.ts
LautaroPetaccio: It could happen if the curator approved the set of items and refreshed immediately, but the contract will fail as it will try to re-use the same cheque. It should be quite uncommon.

nicosantangelo: oh ok, understood!

PR #462, opened 2022-03-24
feat: Update ItemCuration content_hash on TP upsert

juanmahidalgo:
Closes #1899

### DIFF
--- a/src/Item/Item.service.ts
+++ b/src/Item/Item.service.ts
@@ -592,6 +592,13 @@ export class ItemService {
       : null
 
     const upsertedItem: ItemAttributes = await Item.upsert(attributes)
+    if (dbItem && attributes.local_content_hash) {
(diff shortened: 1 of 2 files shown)


src/Item/Item.service.ts
```diff
@@ -592,6 +592,13 @@ export class ItemService {
       : null
 
     const upsertedItem: ItemAttributes = await Item.upsert(attributes)
+    if (dbItem && attributes.local_content_hash) {
+      // Update the Item Curation content_hash
+      await ItemCuration.update(
+        { content_hash: attributes.local_content_hash },
+        { item_id: item.id }
```
LautaroPetaccio, line 599: Would you mind using attributes here?
```suggestion
        { item_id: attributes.id }
```

LautaroPetaccio (approved):

PR #471, opened 2022-03-31
fix: Fix forum post with wrong id

LautaroPetaccio:
This PR fixes an issue where the collection was not being updating properly due to using the id of the post.


juanmahidalgo (approved):
good catch @LautaroPetaccio !! 🚀

PR #491, opened 2022-04-11
feat: Add collectionId query param to the items per address endpoint

juanmahidalgo:
## Why? 

As we introduced pagination for some of the pages in the UI, we faced the need of fetching just the "orphan" items for an address. By adding the `collectionId` query param, we can pass `collectionId=null` and get the items without a collection. 
This will be used in the `/collections` and `/item-editor` pages in the builder UI.

Adds JSDoc attr. to the CollectionFragment 
image

### DIFF
--- a/src/Item/Item.model.ts
+++ b/src/Item/Item.model.ts
@@ -70,7 +70,7 @@ export class Item extends Model<ItemAttributes> {
 
   // PAGINATED QUERIES
 
-  static findAllItemsByAddress(
+  static findItemsByAddress(
     address: string,
     thirdPartyIds: string[],
     parmas: {
--- a/src/Item/Item.model.ts
+++ b/src/Item/Item.model.ts
@@ -71,24 +71,37 @@ export class Item extends Model<ItemAttributes> {
   // PAGINATED QUERIES
 
   static findAllItemsByAddress(
-    thirdPartyIds: string[],
     address: string,
-    limit: number = DEFAULT_LIMIT,
-    offset: number = 0
+    thirdPartyIds: string[],
+    parmas: {
+      collectionId?: string
+      limit?: number
+      offset?: number
+    }
   ) {
+    const { collectionId, limit, offset } = parmas
     return this.query<ItemWithTotalCount>(SQL`
       SELECT items.*, count(*) OVER() AS total_count
         FROM ${raw(this.tableName)} items
-        JOIN ${raw(
+        LEFT JOIN ${raw(
           Collection.tableName
         )} collections ON collections.id = items.collection_id
-        WHERE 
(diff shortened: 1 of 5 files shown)


src/Item/Item.model.ts
```diff
@@ -70,25 +70,38 @@ export class Item extends Model<ItemAttributes> {
...
+      limit?: number
+      offset?: number
+    }
   ) {
+    const { collectionId, limit, offset } = parmas
     return this.query<ItemWithTotalCount>(SQL`
       SELECT items.*, count(*) OVER() AS total_count
         FROM ${raw(this.tableName)} items
-        JOIN ${raw(
+        LEFT JOIN ${raw(
```
LautaroPetaccio, line 86: Would you mind explaining why do we need a LEFT JOIN here? We want all of the returned values to match any of the clauses we're using.

src/Item/Item.router.ts
```diff
@@ -207,12 +211,18 @@ export class ItemRouter extends Router {
...
       allItems
     )
 
     const [items, tpItems] = await Promise.all([
-      Bridge.consolidateItems(dbItems, remoteItems),
+      Bridge.consolidateItems(
+        dbItems,
+        remoteItems.filter((remoteItem) =>
+          collectionIds.includes(remoteItem.collection.id)
+        )
```
LautaroPetaccio, line 224: There shouldn't be any remote items that don't have collections, is there a reason for this filter?

LautaroPetaccio (changes requested):
Would you mind explaining the reason of this PR in the description? As it doesn't have a ticket, it would be interesting to explain the reason why we're adding this parameter.

src/Item/Item.model.ts
juanmahidalgo: The thing is that the `JOIN` statement is joining the tables by `collection.id`.. Without the `LEFT` we wouldn't get the items that don't have a collection associated as the results will be just the ones with one. 

JOIN:
image

LEFT JOIN:
image

src/Item/Item.router.ts
juanmahidalgo: Since we're adding a filter by `collectionId`, at the moment of consolidating the items, we should only use the `remoteItems` assign to that collectionId because later on in `consolidateItems` both `dbItems` and the dbItems from the remoteItems are merged. 
Imagine this case:
```
dbItems = [{ id: 1, collectionId: null} , { id: 2, collectionId: null} ]
remoteItems = [{ id: 3, collectionId: 'id'} ,{ id: 4, collectionId: 'id'}] 
```
if we don't filter the remoteItems, the output of the `consolidateItems` will include the ids `3` and `4`.

LautaroPetaccio (approved):

PR #492, opened 2022-04-12
feat: Return the updated curation in the upsert item endpoint

juanmahidalgo:
Closes #1933

### DIFF
--- a/src/Item/Item.service.ts
+++ b/src/Item/Item.service.ts
@@ -7,7 +7,7 @@ import {
 } from '../Collection'
 import { CollectionService } from '../Collection/Collection.service'
 import { CurationStatus } from '../Curation'
-import { ItemCuration } from '../Curation/ItemCuration'
+import { ItemCuration, ItemCurationAttributes } from '../Curation/ItemCuration'
 import { Bridge } from '../ethereum/api/Bridge'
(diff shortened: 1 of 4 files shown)


src/Item/Item.service.ts
```diff
@@ -637,13 +645,15 @@ export class ItemService {
...
-        { content_hash: attributes.local_content_hash },
-        { item_id: attributes.id, status: CurationStatus.PENDING }
+      curation = await ItemCuration.updateByItemIdAndStatus(
+        attributes.id,
+        CurationStatus.PENDING,
+        { content_hash: attributes.local_content_hash }
       )
     }
-    return Bridge.toFullItem(upsertedItem, dbCollection)
+    return { item: Bridge.toFullItem(upsertedItem, dbCollection), curation }
```
LautaroPetaccio, line 657: This is a breaking change in the API. I think that we should request the curation whenever the item is updated in the UI.

LautaroPetaccio (changes requested):

src/Item/Item.service.ts
juanmahidalgo: Yeah, that's the right call. I've updated https://github.com/decentraland/builder/pull/1968 to fetch the item curation again, so I'm closing this PR.

PR #494, opened 2022-04-13
fix: Items not being fetched with a user that is not a manager of a TP

LautaroPetaccio:
This PR fixes an issue that occurred when requesting the items of a user that wasn't a manager of a third party. The endpoint was returning `[]` for those cases.

### DIFF
--- a/src/Item/Item.service.ts
+++ b/src/Item/Item.service.ts
@@ -172,7 +172,7 @@ export class ItemService {
     }
   ): Promise<(ItemAttributes & { total_count: number })[]> {
     const thirdParties = await thirdPartyAPI.fetchThirdPartiesByManager(address)
-    const thirdPartyIds = thirdParties.map((thirdParty: any) => thirdParty.id)
+    const thirdPartyIds = thirdParties.map((thirdParty) => thirdParty.id)
 
     return Item.findItemsByAddress(address, thirdPartyIds, params)
   }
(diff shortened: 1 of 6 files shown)


src/Item/Item.service.ts
```diff
@@ -172,11 +172,7 @@ export class ItemService {
     }
   ): Promise<(ItemAttributes & { total_count: number })[]> {
     const thirdParties = await thirdPartyAPI.fetchThirdPartiesByManager(address)
-    if (thirdParties.length <= 0) {
-      return []
-    }
-
-    const thirdPartyIds = thirdParties.map((thirdParty) => thirdParty.id)
+    const thirdPartyIds = thirdParties.map((thirdParty: any) => thirdParty.id)
```
nicosantangelo, line 175: why any here :( ?

src/ethereum/api/fragments.ts
```diff
@@ -54,7 +54,6 @@ export const thirdPartyFragment = () => gql`
     root
     managers
     maxItems
-    totalItems
```
nicosantangelo, line 57: Shouldn't we remove it from ThirdPartyFragment too?

LautaroPetaccio: We do! Great catch!

src/Item/Item.service.ts
LautaroPetaccio: I forgot to delete it, it was for testing purposes.

nicosantangelo (approved):

PR #496, opened 2022-04-14
feat: Add specific error for URN duplication cases

### LINKED ISSUE #489: Builder server not reporting the error correctly when updating an item/collection with an already used URN
## 🎉 Description

[Following the fix in the builder server](https://github.com/decentraland/builder/pull/1956) related to creating or updating a collection or an item that has an URN that is being already used, implement a mechanism to return a more detailed error that shows that the creation or update failed do to an URN being already used.

* [x] 🐛 This is a bug report.

juanmahidalgo:
Closes #489

## When creating a collection with an already used URN
image

## When trying to update a collection URN with a one already used
image

## When updating an item and the updated URN is already used

image

### DIFF
--- a/src/Collection/Collection.errors.ts
+++ b/src/Collection/Collection.errors.ts
@@ -29,6 +29,12 @@ export class AlreadyPublishedCollectionError extends Error {
   }
 }
 
+export class URNAlreadyInUseError extends Error {
+  constructor(public id: string, public urn: string, action: CollectionAction) {
+    super(`The selected URN is already being used. It can be ${action}.`)
+  }
+}
+
 export class WrongCollectionError extends Error {
   constructor(m: string, public data: Record<string, any>) {
     super(m)
--- a/src/Collection/Collection.errors.ts
+++ b/src/Collection/Collection.errors.ts
@@ -31,7 +31,9 @@ export class AlreadyPublishedCollectionError extends Error {
 
 export class URNAlreadyInUseError extends Error {
   constructor(public id: string, public urn: string, action: CollectionAction) {
-    super(`The selected URN is already being used. It can be ${action}.`)
+    super(
+      `The URN provided already belongs to a collection. The collection can't be ${action}.`
+    )
   }
 }
 
--- a/src/Item/Item.errors.ts
+++ b/src/Item/Item.errors.ts
@@ -30,7 +30,9 @@ export class ThirdPartyItemAlreadyPublishedError extends Error {
 
 export class URNAlreadyInUseError extends Error {
   constructor(public id: string, public urn: string, action: ItemAction) {
-    super(`The selected URN is already being used. It can be ${action}.`)
+    super(
+      `The URN provided already belong to another item. The item can't be ${action}.`
+    )
   }
 }
 
--- a/src/Item/Item.errors.ts
+++ b/src/Item/Item.errors.ts
@@ -28,6 +28,12 @@ export class ThirdPartyItemAlreadyPublishedError extends Error {
   }
 }
 
+export class URNAlreadyInUseError extends Error {
+  constructor(public id: string, public urn: string, action: ItemAction) {
+    super(`The selected URN is already being used. It can be ${action}.`)
+  }
+}
+
 export class DCLItemAlreadyPublishedError extends Error {
   constructor(
     public id: string,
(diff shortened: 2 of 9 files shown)


src/Item/Item.model.ts
```diff
@@ -68,6 +68,24 @@ export class Item extends Model<ItemAttributes> {
         WHERE ${where}`)
   }
 
+  static async isURNRepeated(
+    id: string,
+    thirdPartyId: string,
+    urnSuffix: string
+  ): Promise<boolean> {
+    const counts = await this.query<{ count: number }>(SQL`
```
LautaroPetaccio, line 76: Count is actually a string, we need to fix other counts too. It's a string because it's a big number.
We also need to take a look at other queries that use count to be sure they work properly.
```suggestion
    const counts = await this.query<{ count: number }>(SQL`
```

src/Item/Item.service.ts
```diff
@@ -645,4 +626,25 @@ export class ItemService {
...
+        id,
+        decodedItemURN.third_party_id,
+        decodedItemURN.item_urn_suffix
+      )
+    ) {
+      throw new URNAlreadyInUseError(id, urn, action)
+    }
+    const [wearable] = await peerAPI.fetchWearables<ThirdPartyWearable>([urn])
+    if (wearable) {
+      throw new URNAlreadyInUseError(id, urn, ItemAction.INSERT)
```
LautaroPetaccio, line 647: Why is it using `ItemAction.INSERT` here?

src/Item/Item.service.ts
```diff
@@ -645,4 +626,25 @@ export class ItemService {
...
+    if (
+      await Item.isURNRepeated(
+        id,
+        decodedItemURN.third_party_id,
+        decodedItemURN.item_urn_suffix
+      )
+    ) {
+      throw new URNAlreadyInUseError(id, urn, action)
+    }
+    const [wearable] = await peerAPI.fetchWearables<ThirdPartyWearable>([urn])
```
LautaroPetaccio, line 645: It might be better to do this call alongside the `Item.isURNRepeated` one with a `Promise.all`.

src/Item/Item.errors.ts
```diff
@@ -28,6 +28,12 @@ export class ThirdPartyItemAlreadyPublishedError extends Error {
   }
 }
 
+export class URNAlreadyInUseError extends Error {
+  constructor(public id: string, public urn: string, action: ItemAction) {
+    super(`The selected URN is already being used. It can be ${action}.`)
```
LautaroPetaccio, line 33: What do you think about this text?
```suggestion
    super(`The provided URN already belong to another item. The item can't be ${action}.`)
```

src/Collection/Collection.errors.ts
```diff
@@ -29,6 +29,12 @@ export class AlreadyPublishedCollectionError extends Error {
   }
 }
 
+export class URNAlreadyInUseError extends Error {
+  constructor(public id: string, public urn: string, action: CollectionAction) {
+    super(`The selected URN is already being used. It can be ${action}.`)
```
LautaroPetaccio, line 34: What do you think about this text?
```suggestion
    super(`The provided URN already belongs to a collection. The collection can be ${action}.`)
```

LautaroPetaccio (changes requested):

src/Item/Item.service.ts
juanmahidalgo: I think calling them in a sequence might be better in some scenarios, where we avoid calling the catalyst if we already know by consulting the database. What do you think?

src/Item/Item.model.ts
juanmahidalgo: You're right, the `count` coming from the query is actually a string. This is just a TS typing issue as they are working properly because JS automatically converts the string to a number when doing the check `counts[0].count >0`
image

I'just changed the type to `string` and converted it to `Number` before the check.

src/Item/Item.service.ts
juanmahidalgo: Nope, just a leftover! thanks!

src/Collection/Collection.errors.ts
juanmahidalgo: That's fine, I believe it should be `the URN provided ...` right?

LautaroPetaccio (approved):

PR #497, opened 2022-04-15
feat: Add item_count on collection upsert

LautaroPetaccio:
This PR returns the `item_count` property when the collections are updated or inserted.

### DIFF
--- a/src/Collection/Collection.model.ts
+++ b/src/Collection/Collection.model.ts
@@ -1,4 +1,5 @@
 import { Model, raw, SQL } from 'decentraland-server'
+import { database } from '../database/database'
 import { Item } from '../Item/Item.model'
 import { CollectionAttributes } from './Collection.types'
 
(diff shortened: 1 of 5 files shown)


src/Collection/Collection.service.ts
```diff
@@ -431,9 +431,9 @@ export class CollectionService {
     }
 
     const attributes = toDBCollection(collectionJSON)
-
     // Should we do something with the salt and the contract address? There's no need to have them
-    return new Collection(attributes).upsert()
+    // return new Collection(attributes).upsert()
```
juanmahidalgo, line 435: Can we remove this comment here?

src/Collection/Collection.model.ts
```diff
@@ -101,4 +101,35 @@ export class Collection extends Model<CollectionAttributes> {
...
+
+  static async upsertWithItemCount(
+    collection: CollectionAttributes
+  ): Promise<CollectionWithItemCount> {
+    const { id, ...attributes } = collection
+    const columnNames = ['id', ...Object.keys(attributes)]
+    const columnValues = [id, ...Object.values(attributes)]
+
+    const insertColumns = columnNames.map((name) => `"${name}"`)
+    const insertValues = columnNames.map((_, index) => `$${index + 1}`)
```
juanmahidalgo, line 113: Can we use the `toAssignmentFields` method as we're doing [here](https://github.com/decentraland/builder-server/blob/54ba34e78b691a5ca789d0c5b8c7cdcb53430cf6/src/Curation/Curation.service.ts#L81)?

src/Collection/Collection.model.ts
```diff
@@ -101,4 +101,35 @@ export class Collection extends Model<CollectionAttributes> {
...
+      .map((name, index) => `${name} = $${index + 2}`)
+
+    const result = await this.query<CollectionWithItemCount>(
+      SQL`
+    INSERT INTO ${raw(this.tableName)} as collections (${raw(
+        insertColumns.join(', ')
+      )})
+      VALUES (${raw(insertValues.join(', '))})
+      ON CONFLICT (id)
+      DO UPDATE SET ${raw(updateFields.join(', '))}
```
juanmahidalgo, line 126: Could we pass directly the array `updateFields` so we avoid the `raw` and `.join(..`?

src/Collection/Collection.model.ts
```diff
@@ -101,4 +101,35 @@ export class Collection extends Model<CollectionAttributes> {
...
+    collection: CollectionAttributes
+  ): Promise<CollectionWithItemCount> {
+    const { id, ...attributes } = collection
+    const columnNames = ['id', ...Object.keys(attributes)]
+    const columnValues = [id, ...Object.values(attributes)]
+
+    const insertColumns = columnNames.map((name) => `"${name}"`)
+    const insertValues = columnNames.map((_, index) => `$${index + 1}`)
+
+    const updateFields = columnNames
```
juanmahidalgo, line 115: Can we use the `toAssignmentFields` method as we're doing [here](https://github.com/decentraland/builder-server/blob/54ba34e78b691a5ca789d0c5b8c7cdcb53430cf6/src/Curation/Curation.service.ts#L81)?

src/Collection/Collection.service.ts
LautaroPetaccio: Removed it!

src/Collection/Collection.model.ts
LautaroPetaccio: Changed it!

juanmahidalgo (approved):

PR #498, opened 2022-04-19
fix: Collection creation

LautaroPetaccio:
This PR fixes the creation of collections that were failing due to a change in the query.


fzavalia (approved):
👌

PR #500, opened 2022-04-20
feat: Rarity endpoints return prices field with prices in various currencies

### LINKED ISSUE #490: /rarities should return converted rarity prices
The endpoint only returns the prices defined in the graph. This was enough when prices were fixed in MANA, but with prices now in USD and paid in MANA, a conversion is required to be done by the endpoint.

fzavalia:
Closes #490


fzavalia:
Requires https://github.com/decentraland/decentraland-transactions/pull/47

src/Rarity/Rarity.router.ts
```diff
@@ -1,18 +1,97 @@
...
+        STATUS_CODES.notFound
+      )
+    }
+
+    return {
+      id: rarity.name,
+      name: rarity.name,
+      price: rarity.price.toString(),
+      maxSupply: rarity.maxSupply.toString(),
+    }
```
LautaroPetaccio, line 95: What if we return the currency of the price. If `inUSD` is not set, it's not clear which is the currency it's returning the price.

src/Rarity/Rarity.router.ts
```diff
@@ -1,18 +1,97 @@
...
 
-  async getRarities() {
+  getRarities = async (req: Request) => {
+    // If the server is still using the old rarities contract.
+    // Return rarities as they have been always returned
+    if (!isUsingRaritiesWithOracle()) {
+      return collectionAPI.fetchRarities()
+    }
+
+    const inUSD = req.query.inUSD === 'true'
```
LautaroPetaccio, line 27: What if we name this query parameter `currency`? We can later have different conversions if we want to.
```suggestion
    const inUSD = req.query.currency === 'usd'
```

src/Rarity/Rarity.router.spec.ts
```diff
@@ -0,0 +1,165 @@
...
+    beforeEach(() => {
+      mockIsUsingRaritiesWithOracle.mockReturnValueOnce(false)
+    })
+
+    it('should fail with an endpoint not found', async () => {
+      await server.get(buildURL('/rarities/common')).expect(404)
+    })
+  })
+
+  // TODO: Tests for when rarities feature flag is enabled
```
LautaroPetaccio, line 164: Tests are missing here :)

LautaroPetaccio (changes requested):

src/Rarity/Rarity.router.ts
fzavalia: I like that!

fzavalia:
@LautaroPetaccio Thanks for the feedback!
- Added missing tests.
- Change query param to currency (defaults to MANA).
- Returned rarities also include a currency field indicating in what currency the price is in.

src/Rarity/Rarity.router.ts
```diff
@@ -1,18 +1,101 @@
...
+    return graphRarities.map((graphRarity) => {
+      // Not handling if the blockchain rarity is not present because that is checked when calling the
+      // this.getRarityFromBlockchain function.
+      const blockchainRarity = blockchainRaritiesMap.get(graphRarity.name)!
+
+      return {
+        ...graphRarity,
+        price: blockchainRarity.price,
+        originalPrice: graphRarity.price,
+        originalCurrency: Currency.USD,
```
LautaroPetaccio, line 49: As "Original" as name might be a bit confusing, what if we change this to a map?

As a map:
```
price: {
'MANA': '4000000000000000000',
'USD': '10000000000000000000'
}
```

LautaroPetaccio (changes requested):

fzavalia:
@LautaroPetaccio Replaced "original" fields with "prices" field containing a Record of Currency -> Price

LautaroPetaccio (approved):

PR #508, opened 2022-04-27
fix: remove isApproved where condition in the query

juanmahidalgo:
Removes the `isApproved: true` since it's not needed anymore.


nicosantangelo (approved):
LGTM

PR #512, opened 2022-05-04
feat: hashV1 migration script

LautaroPetaccio:
Added a script to migrate all items that match one of the following criteria to the hashV1 for its contents:
- Standard items that are part of a published collection (ie. in the blockchain already) but not in the catalyst (ie. a collection just published but not approved yet).
- Standard items that are published but unsynced (ie. the creator made some changes after the collection was approved, it doesn't matter if they have pushed changes or not yet).
- Third party items that are not published.
- Third party items that are published but not in the catalyst (ie. not approved yet).
- Third party items hat are published but are unsynced (ie. the creator made changes to them after they got approved).

The migration script updates the `item.contents` mapping of each item in the DB, uploads a copy of all the files with the new hashV1 to S3, and updates the `item.local_content_hash` to the hashV1 if is different to `null`.

The migration script can be run several times or stopped in the middle of the process and running it again should skip already migrated items in the DB and already copied files in S3.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,3 +10,4 @@ coverage
 !jest.config*.js
 
 .DS_Store
+backup.dump
--- a/.gitignore
+++ b/.gitignore
@@ -11,3 +11,4 @@ coverage
 
 .DS_Store
 backup.dump
+backup*.dump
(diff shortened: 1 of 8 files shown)


LautaroPetaccio:
Closing the script to keep it out of the codebase.

PR #541, opened 2022-06-03
fix: add default order for items by created_at

### LINKED ISSUE #538: Modifying a third party item results in the item being last in the item's pages of the collection
## 🎉 Description

After modifying an item and querying the items of a collection again, the item is placed last in the list of items. The list of items of a collection should be shown in the order that they were created instead of the order on which they were modified. Changing the order of the items might be confusing for the user.

* [x] 🐛 This is a bug report.

juanmahidalgo:
Closes #538

### DIFF
--- a/src/Item/Item.model.ts
+++ b/src/Item/Item.model.ts
@@ -142,34 +142,28 @@ export class Item extends Model<ItemAttributes> {
 
   static async findByCollectionIds(
     collectionIds: string[],
-    status: CurationStatus | undefined,
     limit: number = DEFAULT_LIMIT,
     offset: number = 0
   ) {
     return await this.query<ItemWithTotalCount>(
-      ItemQueries.selectByStandardCollectionIds(
-        collectionIds,
-        status,
-        limit,
-        offset
-      )
+      ItemQueries.selectByCollectionIds(collectionIds, undefined, limit, offset)
     )
   }
 
-  static async findByTPCollectionIdAndStatus(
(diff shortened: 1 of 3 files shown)


src/Item/Item.model.ts
```diff
@@ -181,6 +187,26 @@ export const ItemQueries = Object.freeze({
...
+  ) =>
+    SQL`
+      SELECT items.*, count(*) OVER() AS total_count
+        FROM ${raw(Item.tableName)} items
+        WHERE items.collection_id = ANY(${collectionIds})
+          AND ${
+            status && status === CurationStatus.PENDING
+              ? SQL`items.blockchain_item_id IS NULL`
+              : SQL`1 = 1`
+          }
```
LautaroPetaccio, line 208: The `blockchain_item_id` is independent from the curation status as it's set once the collection is published.
Standard items don't have curation statuses, their collection can have (once it changes and is up to re-review) a curation so I wouldn't mix that concept in the query unless we wanted to get the items of a collection with a pending status.

LautaroPetaccio (changes requested):

src/Item/Item.model.ts
juanmahidalgo: Isn't this that case? The `status` is for the items that are getting asked.
If we're asking for `pending` items, shouldn't we return the ones with `blockchain_item_id` NOT null? I might be missing something here..

PR #543, opened 2022-06-03
feat: also check for blockchain_item_id for isDCLPublished

cazala:
Check first against the DB, if any item has already blockchain_item_id the collection is published (in case both subgraph and RPC are lagging or unresponsive)


.gitignore
```diff
@@ -10,3 +10,5 @@ coverage
 !jest.config*.js
 
 .DS_Store
+.env.migration.*
+backup*
```
fzavalia, line 14: EOL

fzavalia (approved):

src/Collection/Collection.service.ts
```diff
@@ -510,6 +510,20 @@ export class CollectionService {
...
+    })
+    if (!dbCollection) {
+      return false
+    }
+    const collectionItems = await Item.find<ItemAttributes>({
+      collection_id: dbCollection.id,
+    })
+    if (collectionItems.some((item) => item.blockchain_item_id !== null)) {
+      return true
+    }
```
LautaroPetaccio, line 525: What if you create a new query in the `Items.model` file that goes through the items that belong to a collection to see if there's one published? This change will greatly improve the performance of the operation.

LautaroPetaccio (changes requested):

src/Collection/Collection.service.ts
cazala: done ✅

LautaroPetaccio (changes requested):

src/Item/Item.model.ts
```diff
@@ -104,6 +104,22 @@ export class Item extends Model<ItemAttributes> {
     return results[0]
   }
 
+  static async hasPublishedItems(contractAddress: string) {
```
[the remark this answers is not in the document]
cazala: I just pushed a commit that i think it does that, let me know if that's what you meant

LautaroPetaccio (approved):

PR #556, opened 2022-07-05
feat: Filter collections using the request param isPublished

cyaiox:
This PR allows filtering the collections of an address by using the request param: `isPublished`

### DIFF
--- a/src/Collection/Collection.model.ts
+++ b/src/Collection/Collection.model.ts
@@ -122,28 +122,27 @@ export class Collection extends Model<CollectionAttributes> {
   static getPublishedJoinStatement(isPublished: boolean | undefined) {
     return isPublished !== undefined
       ? SQL`
-            JOIN ${raw(Item.tableName)} items ON items.collection_id = c.id ${
-          isPublished
-            ? SQL`AND (
-              (items.blockchain_item_id is NOT NULL AND c.third_party_id IS NULL)
-              OR c.third_party_id is NOT NULL)`
-            : SQL`AND (items.blockchain_item_id is NULL AND c.third_party_id IS NULL)`
-        }
-            LEFT JOIN ${raw(
-              ItemCuration.tableName
-            )} item_curations ON item_curations.item_id = items.id
-            ${
-              isPublished === false
-                ? SQL`
-                  WHERE (
-                    (SELECT COUNT(*) FROM item_curations
-                      LEFT JOIN items on items.id = item_curations.item_id
-                      LEFT JOIN collections cc on cc.id = items.collection_id
-                      WHERE items.collection_id = c.id AND item_curations.item_id = items.id
-                    ) = 0)
-                `
-                : SQL``
-            }
+          LEFT JOIN ${raw(Item.tableName)} items ON items.collection_id = c.id
+          LEFT JOIN ${raw(
+            ItemCuration.tableName
+          )} item_curations ON item_curations.item_id = items.id
+          ${
+            isPublished === false
+              ? SQL`
+                WHERE (
+                  (items.blockchain_item_id is NULL AND c.third_party_id IS NULL) AND
+                  (SELECT COUNT(*) FROM item_curations
+                    LEFT JOIN items on items.id = item_curations.item_id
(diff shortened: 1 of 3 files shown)


src/Collection/Collection.model.ts
```diff
@@ -122,28 +122,27 @@ export class Collection extends Model<CollectionAttributes> {
...
-                  WHERE (
-                    (SELECT COUNT(*) FROM item_curations
-                      LEFT JOIN items on items.id = item_curations.item_id
-                      LEFT JOIN collections cc on cc.id = items.collection_id
-                      WHERE items.collection_id = c.id AND item_curations.item_id = items.id
-                    ) = 0)
-                `
-                : SQL``
-            }
+          LEFT JOIN ${raw(Item.tableName)} items ON items.collection_id = c.id
```
cyaiox, line 125: Modify the Join to a Left Join because the first one excluded the collections without items.

src/Collection/Collection.model.ts
```diff
@@ -122,28 +122,27 @@ export class Collection extends Model<CollectionAttributes> {
...
+          LEFT JOIN ${raw(
+            ItemCuration.tableName
+          )} item_curations ON item_curations.item_id = items.id
+          ${
+            isPublished === false
+              ? SQL`
+                WHERE (
+                  (items.blockchain_item_id is NULL AND c.third_party_id IS NULL) AND
```
cyaiox, line 133: Moved the condition from the `JOIN` for `isPublished: False` to the `WHERE` section

src/Collection/Collection.router.spec.ts
```diff
@@ -904,7 +904,7 @@ describe('Collection router', () => {
         q = 'collection name 1'
         totalCollectionsFromDb = 1
         baseUrl = '/collections'
-        url = `${baseUrl}?limit=${limit}&page=${page}&assignee=${assignee}&status=${status}&sort=${sort}&is_published=${isPublished}&q=${q}`
+        url = `${baseUrl}?limit=${limit}&page=${page}&assignee=${assignee}&status=${status}&sort=${sort}&isPublished=${isPublished}&q=${q}`
```
cyaiox, line 907: Modify the param `is_published` to `isPublished` because the front end sends it in CamelCase format

src/Collection/Collection.router.spec.ts
```diff
@@ -975,59 +975,120 @@ describe('Collection router', () => {
...
-      )
-      ;(ItemCuration.findLastByCollectionId as jest.Mock).mockReturnValueOnce(
-        itemCurationMock
-      )
-      mockThirdPartyCollectionIsPublished(dbTPCollection.id, false)
+      ;(collectionAPI.fetchCollections as jest.Mock).mockResolvedValueOnce([])
       url = `/${wallet.address}/collections`
     })
 
-    it('should return the requested collections with the URN', () => {
```
cyaiox, line 999: Refactor the test to handle better the mocks

src/Collection/Collection.router.ts
```diff
@@ -221,7 +221,7 @@ export class CollectionRouter extends Router {
     req: AuthRequest
   ): Promise<PaginatedResponse<FullCollection> | FullCollection[]> => {
     const { page, limit } = getPaginationParams(req)
-    const { assignee, status, sort, q, is_published } = req.query
+    const { assignee, status, sort, q, isPublished } = req.query
```
cyaiox, line 224: Modify the param is_published to isPublished because the front end sends it in CamelCase format

src/Collection/Collection.router.spec.ts
juanmahidalgo: What do you think if we change the frontend to send it snake case? I'd advocate for that since we have some other query params already as snake case, so we keep using the same standard.

src/Collection/Collection.router.ts
juanmahidalgo: I believe the front is sending `is_published` as snake case in one of the flows. (Check the `toRemoteCollectionQueryParameters` function)

cyaiox: You're right! let me revert these changes

Issue #559, opened 2022-07-07
Create a new type for emotes on the server/db


cazala:
## 🎉 Description

Currently we store all the items under the same schema on the DB. The schema that we use is the one for wearables, which is working now for emotes, but only because we store the minimum data to create an emote. There are certain things that we can't currenty store, for example the play mode, so we always deploy them as `loop: false`. Once we have the UI to select the play mode, there will be no place to store that value on the server. 

## 📝 Details

So we need to create separate types for Wearables and Emotes on the backend instead of the single Item type we currently have. We can take this opporunity to also upgrade from `@dcl/schemas` v4 to v5, which contains already the common types for `Wearable` and `Emote` and use that. We can also remove all these repeated types for [wearables](https://github.com/decentraland/builder-server/blob/master/src/Item/wearable/types.ts) and [emotes](https://github.com/decentraland/builder-server/blob/master/src/Item/emote/types.ts) and simply used the shared types from `@dcl/schemas`. (Note: the `WearableBodyShape` type no longer exists in the common schemas, now it is called `BodyShape`).

PR #560, opened 2022-07-07
feat: use emote entity type

### LINKED ISSUE #558: Use EntityType.EMOTE
## 🎉 Description

Once the platform team introduces the new entity type for emotes, we need to stop deploying all items as wearable entities, use the right type (whether wearable or emote).

## 📝 Details

We need to change the way we compute hashes to match the new schema

cazala:
This PR makes the necessary changes to create the hashes following the new Emote schema.

I created this PR as draft until the platform team enables a catalyst where we can test deploying these new entities.

Closes #558

### DIFF
--- a/src/Item/emote/types.ts
+++ b/src/Item/emote/types.ts
@@ -1,8 +1,21 @@
+import { StandardWearable } from '@dcl/schemas'
 import { WearableBodyShape } from '../wearable/types'
+
+export type Emote = Omit<StandardWearable, 'data'> & {
+  emoteDataADR74: {
+    category: EmoteCategory
+    representations: {
+      bodyShapes: WearableBodyShape[]
+      mainFile: string
+      contents: string[]
+    }[]
+    tags: string[]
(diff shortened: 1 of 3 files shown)


src/Item/emote/types.ts
```diff
@@ -1,8 +1,21 @@
+import { StandardWearable } from '@dcl/schemas'
 import { WearableBodyShape } from '../wearable/types'
 
+export type Emote = Omit<StandardWearable, 'data'> & {
+  emoteDataADR74: {
```
cazala, line 5: Here I copied the schema because we are currently using v4 of `@dcl/schemas` which doesn't have the new emote schema. I'd suggesting upgrading to the new version dhile doing https://github.com/decentraland/builder-server/issues/559 and then delete all the repeated types from this file and this [other one](https://github.com/decentraland/builder-server/blob/master/src/Item/emote/types.ts).

juanmahidalgo (approved):
🚀

PR #564, opened 2022-07-12
feat: Add support for the "synced" parameter in the items collections endpoint

### LINKED ISSUE #563: Add support for 'synced' parameter in the  collection items endpoint
## 🎉 Description

We want to add a way to filter the TPW items by it's synced status. By doing it, it will help TPW admins to manage their collections more efficiently. 

* [ ] 🐛 This is a bug report.
* [X] 📈 This is a feature request.

juanmahidalgo:
Closes #563

### DIFF
--- a/src/Item/Item.model.ts
+++ b/src/Item/Item.model.ts
@@ -204,7 +204,7 @@ export const ItemQueries = Object.freeze({
           SELECT DISTINCT ON (items.id) items.id, items.*
             FROM ${raw(Item.tableName)} items
               ${
-                status || synced
+                status || synced !== undefined
                   ? SQL`
                     JOIN (
                       SELECT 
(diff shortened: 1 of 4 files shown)


src/Item/Item.router.spec.ts
```diff
@@ -515,20 +515,22 @@ describe('Item router', () => {
...
 
-      describe('and pagination params are passed', () => {
+      describe('and pagination and synced params are passed', () => {
         let baseUrl: string
         let page: number
         let limit: number
+        let synced: boolean
         beforeEach(() => {
-          ;(page = 1), (limit = 1)
+          ;(page = 1), (limit = 1), (synced = false)
```
nicosantangelo, line 531: This could be a bit more clear removing the commas right?
```suggestion
          page = 1
          limit = 1
          synced = false
```

src/Item/Item.model.ts
```diff
@@ -190,10 +204,29 @@ export const ItemQueries = Object.freeze({
           SELECT DISTINCT ON (items.id) items.id, items.*
             FROM ${raw(Item.tableName)} items
               ${
-                status
-                  ? SQL`JOIN ${raw(
-                      ItemCuration.tableName
-                    )} item_curations ON items.id = item_curations.item_id`
+                status || synced !== undefined
```
nicosantangelo, line 207: Question, isn't a bit weird that we're receiving the `status` and seemingly not using it? I mean it's used to add a join, but the status value is ignored right?

The concern arise because reading this:
```ts
    const query = ItemQueries.selectByCollectionIds(
      collectionIds,
      undefined,
      synced,
      limit,
      offset
    )
```

was a bit confusing (having `undefined` in the middle like that) but reading the code, it'd been the same if you supplied any Curation status or truthy value, right?

nicosantangelo (approved):

PR #579, opened 2022-08-11
fix: is owned check

juanmahidalgo:


src/Item/Item.service.spec.ts
```diff
@@ -0,0 +1,29 @@
...
+    beforeEach(() => {
+      dbItem = { ...dbItemMock, eth_address: '0xoriginalAddress' }
+      ;(Item.findOne as jest.Mock).mockResolvedValueOnce(dbItem)
+    })
+
+    it('should return true when the owner is the sender', async () => {
+      expect(
+        await service.isOwnedOrManagedBy(dbItem.id, '0xoriginalAddress')
+      ).toBe(true)
+    })
```
LautaroPetaccio, line 21: Would you mind changing the tests to describe the contexes?
```suggestion
    let ownerToCheckFor: string
    describe('when the owner is the same as the one in the DB', () => {
        beforeEach(() => {
           ownerToCheckFor = '0xoriginalAddress'
        })
    
        it('should return true', async () => {
           expect(
             await service.isOwnedOrManagedBy(dbItem.id, ownerToCheckFor)
           ).toBe(true)
        })
    })
```
You can choose to describe the context as a different mock of the dbItem, changing the owner of the item or as changing the owner that is being checked for (as in the example).
The same goes for the test down below.

LautaroPetaccio (changes requested):

src/Item/Item.service.spec.ts
juanmahidalgo: Sure, I can change the test title if it's not descriptive enough. 
Would you mind explaining the benefit of having the variable `ownerToCheckFor`? Does it make any difference?

LautaroPetaccio (approved):

PR #588, opened 2022-08-22
feat: LAND redirection file upload endpoint

fzavalia:


src/LAND/LAND.router.ts
```diff
@@ -0,0 +1,194 @@
...
+      throw new HTTPError(
+        'Failed to upload file to IPFS # Request',
+        {},
+        STATUS_CODES.error
+      )
+    }
+
+    if (!result.ok) {
+      throw new HTTPError(
+        'Failed to upload file to IPFS # Response',
```
LautaroPetaccio, line 68: What if we add more information to the response?
```suggestion
        'Failed to upload file to IPFS # Response, the IPFS server responded with a x error',
```

src/LAND/LAND.router.ts
```diff
@@ -0,0 +1,194 @@
...
+      ipfsHash: Hash,
+      contentHash: await contentHash.fromIpfs(Hash),
+    }
+  }
+
+  private getRedirectionHashes = async (
+    req: Request
+  ): Promise<GetRedirectionHashesResponse> => {
+    const coordsQueryParam = server.extractFromReq(req, 'coords')
+    const separatedCoords = coordsQueryParam.split(';')
```
LautaroPetaccio, line 86: You can send arrays to query parameters by repeating the query parameter:
```
http://localhost:5000?coord=300,100&coord=200,100&coord=500,222
```
We should change it to work that way as it's the HTTP standard.

src/LAND/LAND.router.ts
```diff
@@ -0,0 +1,194 @@
...
+import { getCID } from '../utils/cid'
+import {
+  GetRedirectionHashesResponse,
+  UploadRedirectionResponse,
+} from './LAND.types'
+
+export class LANDRouter extends Router {
+  mount() {
+    this.router.get(
+      '/lands/redirection/hashes',
```
LautaroPetaccio, line 19: I was thinking that we could rename this endpoint to:
```suggestion
      '/lands/redirection-hashes',
```
To be a bit more RESTfullish.

LautaroPetaccio (changes requested):
Great job 🚀 
Would you mind adding tests to the endpoints?

src/LAND/LAND.router.ts
fzavalia: I did it the other way because it meant more characters could be inserted into the URL without hitting the limit.
I like it more the way you propose it still, mostly because the framework parses the query param array for you.
I will change it.

src/LAND/LAND.router.ts
fzavalia: Like the idea.
I would do it camel case like the /assetPacks endpoint

src/LAND/LAND.router.ts
```diff
@@ -0,0 +1,204 @@
...
+    const coordsList =
+      typeof coordsListQP === 'string' ? [coordsListQP] : coordsListQP
+
+    if (coordsList.length > MAX_COORDS) {
+      throw new HTTPError(
+        `Max ${MAX_COORDS} coords`,
+        { amount: coordsList.length },
+        STATUS_CODES.badRequest
+      )
+    }
```
LautaroPetaccio, line 110: I'm not sure if this is necessary taking into consideration the 2048 hard limit of the URL's length. We can still have the max length, but it might be redundant.

LautaroPetaccio (changes requested):

src/LAND/LAND.router.ts
LautaroPetaccio: We're keeping the limit.

src/LAND/LAND.router.ts
fzavalia: Changed the message

src/LAND/LAND.router.ts
```diff
@@ -0,0 +1,194 @@
...
+
+    if (!result.ok) {
+      throw new HTTPError(
+        'Failed to upload file to IPFS # Response',
+        {},
+        STATUS_CODES.error
+      )
+    }
+
+    const { Hash } = await result.json()
```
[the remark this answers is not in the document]
fzavalia: Handling the possibility of an error message

src/LAND/LAND.router.ts
fzavalia: Renamed to /redirectionHashes

LautaroPetaccio (approved):

PR #610, opened 2022-10-07
fix: Duplicated collections when filtering by item tag

cyaiox:
This PR fixes the response of duplicated collections when filtering by items tag


juanmahidalgo (approved):
looks great @cyaiox 👏

PR #612, opened 2022-11-03
feat: Support filter collections by type

### LINKED ISSUE #554: Bug when filtering collections by isPublished status
## 🎉 Description

The `Collection.Service::getCollections` method is returning just 1 record when passing the param: `isPublished`.

* [x] 🐛 This is a bug report.

## 📝 Details

Trying to filter collections by `isPublished` status, the `Collection.Service::getCollections` method is returning only 1 record instead of all the coincidences.

kevinszuchet:
Closes #1724 
Closes #554


src/Collection/Collection.types.ts
```diff
@@ -43,3 +43,8 @@ export type PublishCollectionResponse<T> = {
   items: FullItem[]
   itemCurations?: ItemCurationAttributes[]
 }
+
+export enum CollectionTypeFilter {
+  STANDARD = 'standard',
+  THIRD_PARTY = 'third_party'
+}
```
cyaiox, line 50: np: missing end of line

cyaiox (changes requested):

src/Collection/Collection.model.ts
```diff
@@ -69,16 +70,34 @@ export class Collection extends Model<CollectionAttributes> {
...
+      !assignee &&
+      !status &&
+      !type &&
+      !address &&
+      !thirdPartyIds?.length
+    ) {
       return SQL``
     }
+    const isStandard =
+      'collections.third_party_id is NULL AND collections.urn_suffix is NULL'
```
juanmahidalgo, line 98:
```suggestion
      SQL`collections.third_party_id is NULL AND collections.urn_suffix is NULL`
```
Let's wrap it up with the `SQL` method so we don't have to use the `raw` down below :)

src/Collection/Collection.model.ts
```diff
@@ -91,16 +110,28 @@ export class Collection extends Model<CollectionAttributes> {
         ? [
             CurationStatusFilter.PENDING,
             CurationStatusFilter.APPROVED,
+            CurationStatusFilter.REJECTED,
           ].includes(status)
-          ? SQL`collection_curations.status = ${status} AND collections.contract_address = ANY(${remoteIds})`
-          : status === CurationStatusFilter.REJECTED // Rejected not a single curation in approved
-          ? SQL`collection_curations.status = ${status} AND collections.contract_address = ANY(${remoteIds})`
+          ? SQL`(collection_curations.status = ${status} AND (collections.contract_address = ANY(${remoteIds}) OR ${raw(
```
juanmahidalgo, line 115: I really like what you did to re-use the SQL statements above, I was wondering if we can do the same for 
`collections.contract_address = ANY(${remoteIds}`
maybe we can have one variable like the following and re-use it among the queries
```
const isInRemoteIds = SQL`collections.contract_address = ANY(${remoteIds}`
```

src/Collection/Collection.model.ts
```diff
@@ -91,16 +110,28 @@ export class Collection extends Model<CollectionAttributes> {
...
+            )}) OR (collections.contract_address = ANY(${remoteIds}) AND ${raw(
+              isStandard
+            )}))`
           : status === CurationStatusFilter.UNDER_REVIEW // Under review: isApproved false from the contract OR it's assigned & has pending curation
           ? SQL`collection_curations.assignee is NOT NULL AND collection_curations.status = ${CurationStatusFilter.PENDING}`
           : SQL``
         : undefined,
+      type
+        ? type === CollectionTypeFilter.STANDARD
+          ? SQL`${raw(isStandard)}`
```
juanmahidalgo, line 130:
```suggestion
          ? isStandard
```
these won't be needed anymore if we wrap the variables with the `SQL`

juanmahidalgo (approved):
LGTM! great job 🚀

cyaiox (approved):
🚀

PR #614, opened 2022-11-08
fix: Manager cannot upsert items

### LINKED ISSUE #613: Managers can't upsert items
## 🎉 Description

The builder-server is rejecting managers from upserting items due to [a call to Ownable](https://github.com/decentraland/builder-server/blob/70cf1bc40a9530e1930eb4e9e911d4d89f270a81/src/Item/Item.service.ts#L426) which is only checking if the user owns the item instead of checking if it is also a manager of the collection.

* [x] 🐛 This is a bug report.

cyaiox:
This PR validates when updating a published item that the sender is the owner or a manager.

Closes #613

### DIFF
--- a/src/Item/Item.service.ts
+++ b/src/Item/Item.service.ts
@@ -423,7 +423,25 @@ export class ItemService {
     dbCollection: CollectionAttributes | undefined,
     eth_address: string
   ): Promise<FullItem> {
-    const canUpsert = await new Ownable(Item).canUpsert(item.id, eth_address)
+    let isManager = false
+    const isDbCollectionPublished =
+      dbCollection &&
+      (await this.collectionService.isDCLPublished(
+        dbCollection.contract_address!
+      ))
+
+    if (isDbCollectionPublished) {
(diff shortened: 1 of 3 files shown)


src/Item/Item.service.ts
```diff
@@ -423,7 +423,25 @@ export class ItemService {
...
+    const isDbCollectionPublished =
+      dbCollection &&
+      (await this.collectionService.isDCLPublished(
+        dbCollection.contract_address!
+      ))
+
+    if (isDbCollectionPublished) {
+      const remoteCollection = await collectionAPI.fetchCollection(
+        dbCollection!.contract_address!
+      )
```
LautaroPetaccio, line 436: There's a method (`getCollection`) that does this in the `Collection.service`, what do you think about using it?

LautaroPetaccio (changes requested):

src/Item/Item.service.ts
cyaiox: thanks for the feedback! sure! it'll look better and more readable

LautaroPetaccio (approved):

PR #616, opened 2022-11-21
fix: Encapsulate all SQL ORs between parenthesis

### LINKED ISSUE #416: Filter curations by status
We need to add a new Query String to curation endpoints that allows the caller to filter the result by CurationStatus.

It could look like this:
`?status=pending`

To complete the task:
- [ ] Check if it's necessary to add on both item and collection curations. To start, item curations should be enough but it'll be useful in both
- [ ] Filter the curations by the supplied status only if present. Keep in mind items curations are only getting the last ItemCuration for a collection
- [ ] Validate the status is valid

kevinszuchet:
Closes #416


juanmahidalgo (approved):
LGTM

PR #618, opened 2022-12-09
feat: add possiblity to sort by updated_date and oldest collections

meelrossi:
Add new sorting possiblities to the collections list. Items added: `updated_at_asc`, `updated_at_desc` and `oldest` (this las one is to remain consistent with the `newest` option)

### DIFF
--- a/src/Collection/Collection.router.spec.ts
+++ b/src/Collection/Collection.router.spec.ts
@@ -66,7 +66,7 @@ import {
   SlotUsageCheque,
   SlotUsageChequeAttributes,
 } from '../SlotUsageCheque'
-import { CurationStatus, CurationStatusSort } from '../Curation'
+import { CurationStatus } from '../Curation'
 import { isCommitteeMember } from '../Committee'
 import { app } from '../server'
 import { hasPublicAccess } from './access'
@@ -76,6 +76,7 @@ import {
   CollectionAttributes,
   ThirdPartyCollectionAttributes,
   FullCollection,
+  CollectionSort
 } from './Collection.types'
 
 const server = supertest(app.getApp())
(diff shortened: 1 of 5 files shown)


src/Collection/Collection.router.ts
```diff
@@ -311,7 +311,7 @@ export class CollectionRouter extends Router {
         offset: page && limit ? getOffset(page, limit) : undefined,
         limit,
         address: eth_address,
-        sort: CurationStatusSort.NEWEST,
+        sort: sort as CurationStatusSort || CurationStatusSort.NEWEST,
```
cyaiox, line 314: could you please add a test for this new query param? 🙏

src/Collection/Collection.router.spec.ts
```diff
@@ -1059,12 +1060,15 @@ describe('Collection router', () => {
...
+        totalCollectionsFromDb: number,
+        sort: string
+  
       beforeEach(() => {
         ;(page = 1),
           (limit = 3),
           (isPublished = 'true'),
-          (totalCollectionsFromDb = 1)
+          (totalCollectionsFromDb = 1),
+          (sort=CollectionSort.NAME_ASC)
```
juanmahidalgo, line 1071:
```suggestion
          (sort = CollectionSort.NAME_ASC)
```

src/Collection/Collection.types.ts
```diff
@@ -48,3 +48,13 @@ export enum CollectionTypeFilter {
   STANDARD = 'standard',
   THIRD_PARTY = 'third_party'
 }
+
+export enum CollectionSort {
+  MOST_RELEVANT = 'MOST_RELEVANT',
+  NEWEST = 'NEWEST',
```
juanmahidalgo, line 54: I know this was already named like this, but now that we'll support both newest and oldest, just to keep it consistent with the rest of the enum options, what do you think if we name these ones as `CREATED_AT_DESC` and `CREATED_AT_ASC`? Then the front end will be the one translating newest and latest to `CREATED_AT_DESC` and ``CREATED_AT_ASC`

juanmahidalgo (commented):
I just left two small comments but looks great 🚀

juanmahidalgo (approved):
looks great! feel free to merge whenever the other piece is ready 👍

PR #638, opened 2023-03-21
fix: Make addresses collections endpoint lighter

LautaroPetaccio:
The procedure to get the addresses of collections isn't required to get the remove collections nor to consolidate them as we can deduct that information from the database.
In this PR, the mechanism to get the collections from the graph is removed, alongside the consolidation.


juanmahidalgo (approved):
LGTM!

PR #642, opened 2023-03-27
feat: Add custom max-age and s-maxage headers to /addresses endpoint

juanmahidalgo:
Adds a cache headers to the /addresses endpoint so it gets cached for 5 minutes


kevinszuchet (approved):
LGTM

It would be great to have some tests for this the `getAddressesCollections` handler. At least one that checks the new header was correctly set in the response. WDYT?

juanmahidalgo:
> 

I totally agree but I think this headers will be temporary until we find the real issue, then I'd say we remove the headers to resume the real time responses.

PR #650, opened 2023-04-14
fix: User cannot upsert item

cyaiox:
This PR fixes the item updating when a collaborator previously edited the item.

Fixes: #651


LautaroPetaccio (approved):
Great fix! 🚀

PR #656, opened 2023-05-03
fix: Update owner or manager logic

cyaiox:
This PR updates the upsert item logic to allow owners or managers.

### DIFF
--- a/src/Item/Item.service.ts
+++ b/src/Item/Item.service.ts
@@ -455,11 +455,17 @@ export class ItemService {
         dbCollection.contract_address!
       ))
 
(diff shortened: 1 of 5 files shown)


src/Item/Item.service.ts
```diff
@@ -120,7 +120,7 @@ export class ItemService {
     const dbItem = await Item.findOne<ItemAttributes>(id)
     if (!dbItem) {
       return false
-    } else if (isTPItem(dbItem)) {
+    } else if (dbItem.collection_id) {
```
cyaiox, line 123: Removing `isTPItem` validation, this segment will validate if an `eth_address` is the manager or owner of the `Item`

LautaroPetaccio (approved):
Looks great ✅

PR #659, opened 2023-06-08
fix: take collection creator from graph when upserting item

meelrossi:
Merge collection from graph with the one in the database when upserting an item so that all the validations are done with the latest information of the item

- This is done to fix an issue that a content creator transfer creatorShip to another address using the contract but as the eth_address of the collection in the DB remain the previous one he wasn't able to update the items


juanmahidalgo (approved):
LGTM!

PR #661, opened 2023-06-13
refactor: Enhance the post forum endpoint after publishing a TP collection

kevinszuchet:


src/Forum/Forum.router.ts
```diff
@@ -78,27 +75,13 @@ export class ForumRouter extends Router {
...
-          const { id: postId, link } = await createPost(
-            buildCollectionForumPost(collection, slicedItems)
-          )
-          await Collection.update<CollectionAttributes>(
-            { forum_link: link, forum_id: postId },
-            { id: collectionId }
-          )
-          return link
-        }
+        return this.service.upsertThirdPartyCollectionForumPost(
```
cyaiox, line 79: should we wait for this upsert?

src/Collection/Collection.router.ts
```diff
@@ -400,27 +395,10 @@ export class CollectionRouter extends Router {
...
-              result.collection,
-              result.items.slice(0, MAX_FORUM_ITEMS)
-            )
-          )
-          await Collection.update<CollectionAttributes>(
-            { forum_link: link, forum_id: postId },
-            { id }
-          )
-        }
+        this.forumService.upsertThirdPartyCollectionForumPost(
```
cyaiox, line 398: should we wait for this upsert?

src/Forum/Forum.service.ts
```diff
@@ -0,0 +1,36 @@
...
+    collection: ThirdPartyCollectionAttributes,
+    items: FullItem[]
+  ): Promise<string | undefined> {
+    if (collection.forum_id) {
+      const postData = await getPost(collection.forum_id)
+      await updatePost(
+        collection.forum_id,
+        buildCollectionForumUpdateReply(postData.raw, items)
+      )
+      return
```
cyaiox, line 24: wdyt about returning the link here too?

src/Forum/Forum.router.ts
kevinszuchet: Yes, we should because of the catch. Thanks

src/Collection/Collection.router.ts
kevinszuchet: This was an error on my side. Thanks again!

src/Forum/Forum.service.ts
kevinszuchet: Is it possible to get it with the data that I have here or should I modify the `updatePost` so it returns the same as `createPost` and we can get the link there?

cyaiox: I like how you updated the `updatePost` method to return the same link. sorry for the late reply

cyaiox (approved):
amazing job! 🚀

PR #690, opened 2023-08-03
feat: add contents endpoint for preview

meelrossi:
Add `/contents` endpoint that returns the data depending on the content file request

`preview` will return project scene entity object
`composite.main(hash)` will return scene's composite
`hash` will redirect to s3 bucket

### DIFF
--- a/src/Project/Project.router.ts
+++ b/src/Project/Project.router.ts
@@ -17,7 +17,7 @@ import {
 import { withAuthentication, AuthRequest } from '../middleware/authentication'
 import { Ownable } from '../Ownable'
 import { SDK7Scene } from '../Scene/SDK7Scene'
-import { COMPOSITE_FILE_HASH } from '../Scene/composite'
+import { COMPOSITE_FILE_HASH, PREVIEW_HASH } from '../Scene/utils'
 import { Project } from './Project.model'
 import { ProjectAttributes, projectSchema } from './Project.types'
 import { SearchableProject } from './SearchableProject'
@@ -271,7 +271,7 @@ export class ProjectRouter extends Router {
     const content = server.extractFromReq(req, 'content')
 
     // when content is preview, return entity object
-    if (content === 'preview') {
+    if (content === PREVIEW_HASH) {
       try {
         const { scene, project } = await getProjectManifest(projectId)
         if (scene.sdk6) {
--- a/src/Project/Project.router.ts
+++ b/src/Project/Project.router.ts
@@ -39,7 +39,7 @@ const FILE_NAMES = [
 const MIME_TYPES = ['image/png', 'image/jpeg']
 
 const validator = getValidator()
-const scenePreviewMain = fs.readFileSync('src/Project/scene-preview.js')
+const scenePreviewMain = fs.readFileSync('static/scene-preview.js.raw', 'utf-8')
 export class ProjectRouter extends Router {
   mount() {
     const withProjectExists = withModelExists(Project, 'id', {
@@ -304,7 +304,7 @@ export class ProjectRouter extends Router {
     const content = server.extractFromReq(req, 'content')
 
     if (content === INDEX_HASH) {
-      return res.send(new TextDecoder().decode(scenePreviewMain))
+      return res.send(scenePreviewMain)
     }
 
     // when content is preview, return entity object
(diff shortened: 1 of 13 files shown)


src/Project/Project.router.ts
```diff
@@ -228,8 +235,8 @@ export class ProjectRouter extends Router {
 
     addInmutableCacheControlHeader(res)
     return res.redirect(
-      `${getBucketURL()}/${project.getFileKey(filename)}`,
-      301
+      301,
+      `${getBucketURL()}/${project.getFileKey(filename)}`
```
cazala, line 239: were these params in the wrong order all this time?

src/Project/Project.router.ts
```diff
@@ -258,4 +265,70 @@ export class ProjectRouter extends Router {
...
+                "Can't preview projects with sdk6 scene"
+              )
+            )
+        }
+
+        const entity = await new SDK7Scene(scene.sdk7).getEntity(project)
+
+        // Add composite file
+        entity.content = [
+          { file: 'composite.json', hash: COMPOSITE_FILE_HASH },
```
cazala, line 292: This should be `main.composite`

cazala (changes requested):

src/Project/Project.router.ts
meelrossi: It accepts both, but the one we had before was "deprecated"

```
    redirect(status: number, url: string): void;
    redirect(url: string, status: number): void;
```

src/Scene/SDK7Scene.ts
```diff
@@ -28,7 +28,7 @@ export class SDK7Scene {
     const base = this.getLayout().base
     return {
       owner: '',
-      main: 'bin/game.js',
```
cazala: is this okay or in the end we should actually point to a `.js` file?

src/Project/Project.router.ts
```diff
@@ -290,6 +300,62 @@ export class ProjectRouter extends Router {
...
+            )
+        }
+
+        const entity = await new SDK7Scene(scene.sdk7).getEntity(project)
+
+        // Add composite file
+        entity.content = [
+          { file: "bin/index.js", hash: INDEX_HASH },
+          { file: "main.crdt", hash: CRDT_HASH },
+          ...entity.content.filter(({ file }) => file !== 'main.crdt'),
```
cazala, line 332: is this necessary now that the `main.crdt` is stored separately?

meelrossi: I added it because the scene i was testing had it in the mappings and forgot to remove it. But no, its not necessary

src/Project/Project.router.ts
```diff
@@ -290,6 +299,62 @@ export class ProjectRouter extends Router {
     return true
   }
 
+  async getContents(req: Request, res: Response) {
+    const projectId = server.extractFromReq(req, 'id')
+    const content = server.extractFromReq(req, 'content')
+
+    if (content === INDEX_HASH) {
+      return res.send(new TextDecoder().decode(scenePreviewMain))
```
cazala, line 307: we could avoid using a `TextDecoder` here by passing `utf-8` as the second argument of `readFileSync`, ie:

```
const stringValue = fs.readFileSync('file.js', 'utf-8') 
```

src/Project/Project.router.ts
```diff
@@ -34,6 +39,7 @@ const FILE_NAMES = [
 const MIME_TYPES = ['image/png', 'image/jpeg']
 
 const validator = getValidator()
+const scenePreviewMain = fs.readFileSync('src/Project/scene-preview.js')
```
cazala, line 42: should we put the `scene-prevew` in a separate folder? to me it feels a bit off for this .js file to be here in the `Project` directory.

Maybe instead of `src/Project/scene-preview.js` place it on  `static/scene-preview.js` or something like that?

cazala (approved):

src/Project/Project.router.ts
```diff
@@ -290,6 +299,62 @@ export class ProjectRouter extends Router {
...
+        return res
+          .status(STATUS_CODES.notFound)
+          .send(server.sendError({ projectId }, (error as Error)?.message))
+      }
+    }
+
+    // when content is crdt, return scene crdt file
+    if (content === CRDT_HASH) {
+      const redirectPath = `${getBucketURL()}/${new S3Project(projectId).getFileKey(CRDT_FILENAME)}`
+      console.log({ redirectPath })
```
cazala, line 345: oops

PR #696, opened 2023-08-21
fix: Update SW video field

cyaiox:
This PR updates the video field for smart wearables when creating a new curation.

### DIFF
--- a/src/Curation/Curation.router.spec.ts
+++ b/src/Curation/Curation.router.spec.ts
@@ -22,6 +22,7 @@ import { ItemCuration, ItemCurationAttributes } from './ItemCuration'
 import { CurationService } from './Curation.service'
 import { CurationStatus } from './Curation.types'
 import { createAssigneeEventPost, ForumNewPost } from '../Forum'
+import { VIDEO_PATH } from '../Item/utils'
 
 jest.mock('../common/Router')
 jest.mock('../common/ExpressApp')
@@ -962,6 +963,9 @@ describe('when handling a request', () => {
 
       describe('when updating a collection curation', () => {
         let expectedCuration: CollectionCurationAttributes
+        let itemSpy: jest.SpyInstance<
+          ReturnType<typeof Item['findByCollectionIds']>
+        >
 
         beforeEach(() => {
           expectedCuration = {
@@ -971,6 +975,9 @@ describe('when handling a request', () => {
           service = mockServiceWithAccess(CollectionCuration, true)
           mockIsCommitteeMember.mockResolvedValueOnce(true)
           jest.spyOn(service, 'getLatestById').mockResolvedValueOnce(undefined)
+          itemSpy = jest
+            .spyOn(Item, 'findByCollectionIds')
+            .mockResolvedValueOnce([{ ...dbItemMock, total_count: 1 }])
         })
 
         it('should resolve with the inserted curation', async () => {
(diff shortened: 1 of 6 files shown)


src/Curation/Curation.router.ts
```diff
@@ -498,6 +503,15 @@ export class CurationRouter extends Router {
     return dbItem.local_content_hash
   }
 
+  /* This method updates the video field of smart wearables
+   * after the collection curation is approved.
+   * This way we can handle if the video was updated after the collection was published
+   */
+  private updateCollectionItemsContent = async (collectionId: string) => {
+    const itemService = new ItemService()
+    itemService.updateDCLItemsContent(collectionId)
```
kevinszuchet, line 512: Should we add the `await` here?

cyaiox: you're right! added the `await`

src/Curation/Curation.router.spec.ts
```diff
@@ -594,6 +595,115 @@ describe('when handling a request', () => {
...
+              it('should resolved with the inserted curation and update the item video', async () => {
+                updateSpy = jest
+                  .spyOn(service, 'updateById')
+                  .mockResolvedValueOnce(expectedCuration)
+                const updateItemSpy = jest
+                  .spyOn(Item, 'upsert')
+                  .mockResolvedValue({
+                    ...mockItem,
+                    video: newVideoHash,
+                  })
```
LautaroPetaccio, line 661: Would you mind setting these two in the `beforeEach`?
The same goes for the other test.

src/Item/Item.service.spec.ts
```diff
@@ -98,6 +102,57 @@ describe('Item Service', () => {
...
+            ...dbItemMock,
+            eth_address: oldCreatorAddress,
+            contents: {
+              ...dbItemMock.contents,
+              'male/game.js': 'hash1',
+              [VIDEO_PATH]: 'videoHash2',
+            },
+            video: 'videoHash',
+          }
+          ;(Item.findOne as jest.Mock).mockRestore()
```
LautaroPetaccio, line 124: Did you need to use restores because there was an issue with the mocks?

src/Item/Item.service.spec.ts
```diff
@@ -98,6 +102,57 @@ describe('Item Service', () => {
...
+          ;(Collection.findByIds as jest.Mock).mockImplementation((ids) =>
+            [dbCollection].filter((collection) => ids.includes(collection.id))
+          )
+          ;(collectionAPI.fetchCollection as jest.Mock).mockRestore()
+          ;(collectionAPI.fetchCollection as jest.Mock).mockImplementation(() =>
+            Promise.resolve({
+              ...collectionFragmentMock,
+              creator: oldCreatorAddress,
+            })
+          )
```
LautaroPetaccio, line 138: The mocks that return values can be simplified using the `resolve` or `return` mocks. Would you mind changing them?
```suggestion
          ;(collectionAPI.fetchCollection as jest.Mock).mockResolvedValueOnce({
              ...collectionFragmentMock,
              creator: oldCreatorAddress,
            })
```

src/Item/Item.service.spec.ts
```diff
@@ -98,6 +102,57 @@ describe('Item Service', () => {
...
+              ...dbItemMock.contents,
+              'male/game.js': 'hash1',
+              [VIDEO_PATH]: 'videoHash2',
+            },
+            video: 'videoHash',
+          }
+          ;(Item.findOne as jest.Mock).mockRestore()
+          ;(Item.findOne as jest.Mock).mockResolvedValueOnce(dbItem)
+          ;(Item.hasPublishedItems as jest.Mock).mockResolvedValue(true)
+          ;(Item.upsert as jest.Mock).mockImplementation((value) => value)
```
LautaroPetaccio, line 127: Would you mind using the `once` version of the mocks? This will prevent the mocks from leaking into other tests.

LautaroPetaccio (commented):
The changes look great 👍 
I've left some comments on the tests that I think we could improve.

src/Item/Item.service.spec.ts
cyaiox: not required the restores, removing it

LautaroPetaccio (approved):
🚀

PR #699, opened 2023-08-29
fix: use verify from @dcl/platform-crypto-middleware

kuruk-mm:


src/middleware/authentication.ts
```diff
@@ -60,35 +62,26 @@ async function decodeAuthChain(req: Request): Promise<string> {
...
+        await verify(req.method, req.path, req.headers, {
+          fetcher: peerAPI.signatureFetcher
+        })
       } catch (error) {
         errorMessage = isErrorWithMessage(error) ? error.message : 'Unknown'
+        try {
+          await validateSignature(req, authChain)
+          errorMessage = null // clear error if it has success
+        } catch (_) {
+          // we use the error reported
```
juanmahidalgo, line 83: Should we override the `errorMessage` variable here? The `validateSignature` throws an error that might be different from the one that is thrown by the `verify` function
```suggestion
      catch (error) {
           errorMessage = isErrorWithMessage(error) ? error.message : 'Unknown'
       }
```

juanmahidalgo (approved):
LGTM!! thanks for doing this 👏

PR #710, opened 2023-09-19
feat: Add support for fuzzy search for /collections endpoint

### LINKED ISSUE #709: Provide an endpoint to search collections by name
## 🎉 Description

As the title describes, provide a public endpoint to search across collection names with a fuzzy mecanism to prevent typos and errors.

* [ ] 🐛 This is a bug report.
* [X] 📈 This is a feature request.

juanmahidalgo:
This PR does the following:
- Add a migration to add `pg_trgm` extension and set the `similarity_threshold` to 0.5 (0.3 is the default)
- Allows calls to `/collections` if it's being asked with `q` and `isPublished=true` parameters
- Adds the `%` operator to do the fuzzy search in the Collections query.

Closes #709

### DIFF
--- a/migrations/1695119925617_add-trgm-extension.ts
+++ b/migrations/1695119925617_add-trgm-extension.ts
@@ -1,0 +1,13 @@
+/* eslint-disable @typescript-eslint/naming-convention */
+import { MigrationBuilder, ColumnDefinitions } from 'node-pg-migrate'
+
+export const shorthands: ColumnDefinitions | undefined = undefined
+
+export async function up(pgm: MigrationBuilder): Promise<void> {
+  pgm.addExtension('pg_trgm', { ifNotExists: true })
+  pgm.sql('SET pg_trgm.similarity_threshold = 0.5')
+}
+
+export async function down(pgm: MigrationBuilder): Promise<void> {
+  pgm.dropExtension('pg_trgm')
+}
--- a/src/Collection/Collection.router.spec.ts
+++ b/src/Collection/Collection.router.spec.ts
@@ -849,7 +849,7 @@ describe('Collection router', () => {
       })
     })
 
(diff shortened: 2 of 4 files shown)


migrations/1695119925617_add-trgm-extension.ts
```diff
@@ -0,0 +1,13 @@
+/* eslint-disable @typescript-eslint/naming-convention */
+import { MigrationBuilder, ColumnDefinitions } from 'node-pg-migrate'
+
+export const shorthands: ColumnDefinitions | undefined = undefined
+
+export async function up(pgm: MigrationBuilder): Promise<void> {
+  pgm.addExtension('pg_trgm', { ifNotExists: true })
```
LautaroPetaccio, line 7: I've tried before to install extensions using migrations with no luck. It is pretty possible that this will fail when deploying the service in our infra so it might be better to do it manually.

src/Collection/Collection.router.spec.ts
```diff
@@ -828,222 +828,466 @@ describe('Collection router', () => {
...
-        limit: number,
-        baseUrl: string,
-        totalCollectionsFromDb: number,
-        q: string,
-        assignee: string,
-        status: string,
-        type: string,
-        sort: string,
-        isPublished: string
+    describe('and its not a committee member but has request has the params "q" for a search term and "isPublished"', () => {
```
LautaroPetaccio, line 852:
```suggestion
    describe('and its not a committee member but the request has the params "q" for a search term and "isPublished"', () => {
```

LautaroPetaccio (commented):
The PR looks great! We might want to run the query in the infra instead of using it in a migration.

LautaroPetaccio (approved):

PR #721, opened 2023-11-09
fix: remove the utm_campaign field from subscriptions

sandrade-dcl:
Since the `utm_campaign` field is not adding any value and, in addition, is generating confusion now that we're using the subscriptions endpoints from different sources (builder, explorer...), we've decided to remove it.


juanmahidalgo (approved):
LGTM!

PR #735, opened 2024-04-01
fix: Catch errors when uploading files

LautaroPetaccio:
This PR fixes an error that crashed the server whenever a file upload throws. Catching the error and calling the callback function results in the multer middleware handling the error correctly.


juanmahidalgo (approved):
LGTM!

meelrossi (approved):

PR #738, opened 2024-04-04
feat: add validation for the amount the tags being added

juanmahidalgo:


src/Item/Item.service.ts
```diff
@@ -73,6 +74,16 @@ export class ItemService {
         )
       : await Item.findOne<ItemAttributes>(item.id)
 
+    if (item.data.tags.length > MAX_TAGS_LENGTH) {
+      const isAlreadyExcided =
```
LautaroPetaccio, line 78:
```suggestion
      const isAlreadyExceeded =
```

src/Item/Item.service.spec.ts
```diff
@@ -123,6 +126,123 @@ describe('Item Service', () => {
...
+        })
+      })
+
+      describe('and it is an update operation', () => {
+        beforeEach(() => {
+          ;(Item.findByURNSuffix as jest.Mock).mockResolvedValueOnce(dbItem)
+        })
+        describe('and the item already has the maximun amount of tags', () => {
+          it('should throw the MaximunAmountOfTagsReachedError error', () => {
+            return expect(() =>
```
LautaroPetaccio, line 201: The rejects modifier expects a promise, so It shouldn't be necessary to wrap the expectation on a function, would you mind removing it?
```suggestion
            return expect(
```

LautaroPetaccio (approved):

PR #741, opened 2024-04-22
feat: Add granular CORS

LautaroPetaccio:
This PR removes the global CORS configuration and moves it to each router to granularly set CORS.
Not only the paths must be processed by the CORS middleware, but the OPTIONS headers as well. There are request that wouldn't require an OPTIONS handler, but the Builder send some request with the authentication header, making a preflight request which should be handled.

### DIFF
--- a/src/Curation/Curation.router.ts
+++ b/src/Curation/Curation.router.ts
@@ -4,6 +4,7 @@ import { Router } from '../common/Router'
 import { HTTPError, STATUS_CODES } from '../common/HTTPError'
 import { withAuthentication, AuthRequest } from '../middleware'
 import { isCommitteeMember } from '../Committee'
+import { withCors } from '../middleware/cors'
 import { collectionAPI } from '../ethereum/api/collection'
 import { thirdPartyAPI } from '../ethereum/api/thirdParty'
 import { getValidator } from '../utils/validator'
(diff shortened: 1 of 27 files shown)


src/Curation/Curation.router.ts
```diff
@@ -40,38 +41,55 @@ export class CurationRouter extends Router {
...
     //   - etc
+
+    /**
+     * CORS for the OPTIONS header
+     */
+    this.router.options('/curations', withCors)
+    this.router.options('/collectionCuration/:id/itemsStats', withCors)
+    this.router.options('/collections/:id/itemCurations', withCors)
+    this.router.options('/collections/:id/curation', withCors)
+    this.router.options('/collections/:id/curation/post', withCors)
```
cyaiox, line 52: missing cors for: 
```
this.router.post(
  '/collections/:id/curation/post',
  withAuthentication,
  server.handleRequest(this.createCurationNewAssigneePost)
)
```

src/ThirdParty/ThirdParty.router.ts
```diff
@@ -32,6 +41,7 @@ export class ThirdPartyRouter extends Router {
      */
     this.router.get(
       '/thirdParties/:id',
+      withCors,
```
cyaiox, line 44: should we add: `this.router.options('/thirdParties/:id', withCors)`?

src/Curation/Curation.router.ts
LautaroPetaccio: Good catch, added!

src/ThirdParty/ThirdParty.router.ts
LautaroPetaccio: Good catch, added!

cyaiox (approved):

PR #764, opened 2024-08-14
feat: Return full updated items after upsert

LautaroPetaccio:
This PR changes the upsert endpoint to return the full item after upserting it.

### DIFF
--- a/src/Item/Item.service.ts
+++ b/src/Item/Item.service.ts
@@ -355,10 +355,12 @@ export class ItemService {
   }
 
   private async getTPItem(
-    dbItem: ThirdPartyItemAttributes
+    dbItem: ThirdPartyItemAttributes,
+    dbCollection?: CollectionAttributes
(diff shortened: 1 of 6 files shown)


src/Item/Item.service.ts
```diff
@@ -375,24 +377,27 @@ export class ItemService {
         : collection
 
       const catalystItems = await peerAPI.fetchWearables<Wearable>([urn])
-      if (catalystItems.length > 0) {
-        item = Bridge.mergeTPItem(dbItem, collection, catalystItems[0])
-      }
+      item = Bridge.mergeTPItem(
```
cyaiox, line 380: do you think this validation should be kept? 

`catalystItems.length > 0`

LautaroPetaccio: If there's no catalyst item, the `mergeTPItem` function will receive it as undefined, producing a full item without it.

cyaiox (approved):

PR #772, opened 2024-09-11
feat: Add virtual third parties

LautaroPetaccio:
This PR adds the capability of using creating virtual third parties. That is, third parties created with the purpose of being temporary third parties until the registration is done on chain.


src/Item/Item.service.ts
```diff
@@ -11,7 +11,7 @@ import { ItemCuration } from '../Curation/ItemCuration'
 import { Bridge } from '../ethereum/api/Bridge'
 import { collectionAPI } from '../ethereum/api/collection'
 import { peerAPI } from '../ethereum/api/peer'
-import { thirdPartyAPI } from '../ethereum/api/thirdParty'
+// import { thirdPartyAPI } from '../ethereum/api/thirdParty'
```
cyaiox, line 14:
```suggestion

```

src/Item/Item.service.ts
```diff
@@ -49,6 +49,7 @@ import {
   isTPItem,
   toDBItem,
 } from './utils'
+import { ThirdPartyService } from '../ThirdParty/ThirdParty.service'
```
cyaiox, line 52: could you move this import above?

migrations/1725996635075_add-virtual-third-parties.ts
```diff
@@ -0,0 +1,110 @@
...
+  // Create the trigger to update the updated_at column
+  pgm.createTrigger(
+    VirtualThirdParty.tableName,
+    update_updated_at_trigger_name,
+    {
+      when: 'BEFORE',
+      operation: 'UPDATE',
+      level: 'ROW',
+      function: update_updated_at_column_function_name,
+      condition: 'OLD.* IS DISTINCT FROM NEW.*',
```
cyaiox, line 65: shouldn't the updated_at field be updated with each update? or should it meet a specific condition?

src/ThirdParty/ThirdParty.service.ts
```diff
@@ -15,18 +65,74 @@ export class ThirdPartyService {
...
+
+  static async getThirdParties(manager?: string): Promise<ThirdParty[]> {
+    const [fragments, virtualThirdParties] = await Promise.all([
+      thirdPartyAPI.fetchThirdPartiesByManager(manager),
+      manager
+        ? this.getVirtualThirdPartiesByManager(manager)
+        : ([] as ThirdParty[]),
+    ])
+    const graphThirdParties = fragments.map(toThirdParty)
+    console.log({ graphThirdParties, virtualThirdParties })
```
cyaiox, line 78:
```suggestion
```

migrations/1725996635075_add-virtual-third-parties.ts
LautaroPetaccio: The `updated_at` column will get updated only if any column in the row gets updated to something new. If it doesn't, we'll be saving a write operation by not updating the `updated_at`.

cyaiox (approved):

PR #774, opened 2024-09-23
feat: Add programmatic third parties

LautaroPetaccio:
This PR adds the isProgrammatic property to the third parties, either by consuming the property from the graph or from the virtual third party.

### DIFF
--- a/src/ThirdParty/ThirdParty.service.ts
+++ b/src/ThirdParty/ThirdParty.service.ts
@@ -19,7 +19,8 @@ export class ThirdPartyService {
   static async createVirtualThirdParty(
     id: string,
     managers: string[],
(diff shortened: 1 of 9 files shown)


src/ThirdParty/ThirdParty.service.ts
```diff
@@ -31,6 +32,7 @@ export class ThirdPartyService {
       id,
       managers,
       raw_metadata: raw_metadata,
+      isProgrammatic: isProgrammatic ?? false,
```
cyaiox, line 35: np: 
```suggestion
      isProgrammatic: !!isProgrammatic,
```

cyaiox (approved):

PR #798, opened 2026-03-11
fix: AssetPacks router

cyaiox:
This PR fixes the upserting of assets' parent IDs to another parent.

Additionally, adds auth to the assets endpoint.

### DIFF
--- a/src/Asset/Asset.model.ts
+++ b/src/Asset/Asset.model.ts
@@ -44,10 +44,25 @@ export class Asset extends Model<AssetAttributes> {
         AND ap.is_deleted = FALSE`)
     } else {
       query.append(SQL`
-      WHERE a.id = ANY(${ids})`)
+      INNER JOIN ${SQL.raw(AssetPack.tableName)} ap ON a.asset_pack_id = ap.id
+      WHERE a.id = ANY(${ids})
+        AND ap.is_deleted = FALSE`)
     }
 
     return this.query<AssetAttributes>(query)
+  }
+
+  static async existsAnyWithADifferentAssetPackId(
+    ids: string[],
+    assetPackId: string
+  ): Promise<boolean> {
+    const counts = await this.query(SQL`
+    SELECT COUNT(*) as count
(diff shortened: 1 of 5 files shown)


src/AssetPack/AssetPack.router.ts
```diff
@@ -315,7 +315,11 @@ export class AssetPackRouter extends Router {
     }
 
     const upsertResult = await new AssetPack(attributes).upsert()
-    await Promise.all(assets.map((asset) => new Asset(asset).upsert()))
+    await Promise.all(
+      assets.map((asset) =>
+        new Asset({ ...asset, asset_pack_id: id }).upsert()
```
LautaroPetaccio: What if, instead of rewriting the assets with the asset pack id from the request, we return a 400 if one of them does not have the correct id?

LautaroPetaccio (approved):

src/Asset/Asset.model.ts
```diff
@@ -32,11 +32,22 @@ export class Asset extends Model<AssetAttributes> {
...
+  static findByIds(ids: string[], ethAddresses?: string[]) {
+    const query = SQL`
+    SELECT a.* FROM ${SQL.raw(this.tableName)} a`
+
+    if (ethAddresses) {
+      query.append(SQL`
+      INNER JOIN ${SQL.raw(AssetPack.tableName)} ap ON a.asset_pack_id = ap.id
+      WHERE a.id = ANY(${ids})
+        AND ap.eth_address = ANY(${ethAddresses})
+        AND ap.is_deleted = FALSE`)
```
LautaroPetaccio, line 44: Why do we need the is_deleted condition here and not in the original query?

src/AssetPack/AssetPack.router.ts
cyaiox: good idea!

src/Asset/Asset.model.ts
cyaiox: good catch, will add it to the original query, as we don't want to show deleted assets even using the id.

PR #809, opened 2026-08-20
fix(auth): upgrade to crypto-middleware 6 and refuse non-canonical signers

LautaroPetaccio:
## What this adds on top of #808

Rebased onto master, which now contains **#808** — that PR already migrated off `@dcl/platform-crypto-middleware` and added `declaresSceneSigner` to guard the legacy signature fallback. This PR is now much smaller than when it was opened: two things remained.

**1. `^5.1.0` → `^6.2.0`.** Up to 5.x the signed payload folded the whole joined string, so metadata casing fell outside the signature and two requests differing only in case shared one valid signature. 6.0.0 joins the metadata bytes verbatim, so the parsed metadata `verify()` returns is pinned to what was signed.

**2. The post-verify check moves to `rejectIfSigner`:**

```diff
- if (data.authMetadata.signer === SCENE_SIGNER) {
+ if (!isNotSceneSigner(data.authMetadata)) {
```

This covers what byte binding cannot. A client that *signs* `Decentraland-Kernel-Scene` itself produces a genuinely valid signature — no signature check can refuse it, and only a gate can. The exact comparison returned `false` and the request was accepted as directly user-signed. `rejectIfSigner` refuses a `signer` that is not already canonical rather than comparing it, and nothing is folded: the value reaching handlers is what was signed.

## `declaresSceneSigner` is deliberately left alone

It reads the raw header and normalizes with `trim().toLowerCase()` before comparing, which looks like the pattern this migration has been removing elsewhere. It is correct here, and the difference is its position: it runs **before** `verify()` specifically to cover the fallback path, where a `method:path` signature binds no metadata at all. With nothing pinning that value, the check has to be maximally suspicious rather than exact — normalizing to catch *more* scene requests is defensive, not the fold-then-compare that created the original gap.

That placement is also better than what this PR originally proposed. My first version kept the check only on `verify()`'s success path, which left the fallback ungated; #808's approach covers both.

## Tests

`authentication.spec.ts` previously did a bare `jest.mock('@dcl/crypto-middleware')`, which auto-mocks every export. Left as-is, `rejectIfSigner` would return `undefined`, so the predicate built at import time would be undefined and the scene check would throw rather than run. It now mocks only `verify`, so the tests exercise the real comparison.

Added coverage for the spellings the exact match let through — re-cased, upper-cased, whitespace-padded — each asserting the request is refused.

32 tests passing across `src/middleware`. `tsc` clean, tslint clean on both changed files.

## Breaking change

Callers still signing the pre-6.0.0 payload are rejected whenever their metadata contains an uppercase character. The known callers — the builder dapp, `sites`, `landing` — send empty or all-lowercase metadata (`{}`, `{type: 'allow-list'}`), byte-identical under both formats, so they are unaffected and no legacy fallback is enabled here.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


LautaroPetaccio:
Bumped `@dcl/crypto-middleware` to `^6.3.0` and refreshed the lockfile (the caret range already allowed 6.3.0, but the lockfile pinned 6.2.0, so the fix would not have arrived without the refresh). Lockfile diff is only the `crypto-middleware` entry — no unrelated churn, and its own dependency set is unchanged.

**Why 6.3.0 matters here.** It adds a `hasFoldedVariant` guard to `canonicalField` / `requireCanonicalField`, which `rejectIfSigner` is built on: a key that case-folds to the declared field without being spelled exactly that is now a **rejection** rather than an absence.

That lands on this PR specifically because `declaresRefusedMetadata()` delegates its signer decision to `isNotSceneSigner = rejectIfSigner(SCENE_SIGNER)` instead of comparing by hand. Up to 6.2.0 that delegation was also how the hole reached this gate — the predicate reads the exact `signer` key, so metadata delivering:

```json
{"Signer":"decentraland-kernel-scene"}
```

presented no `signer` at all, the gate read it as "not claiming to be a scene", and `decodeAuthChain` **resolved to the owner address** — a scene-signed request authenticating as the visitor, on metadata that visibly names the signer the gate exists to refuse.

It is reachable on the strict (current-format) path rather than only through the legacy fallback: re-spelling the key changes the signed bytes, so the request has to be signed that way — which a malicious scene-driven client can simply do.

Worth stating plainly since it was the open question on this branch: **the local gate did not need its own fix.** Because it delegates rather than reading `metadata.signer` itself, it inherits the library's guard. No change to `authentication.ts` in this commit.

**Regression test + counterfactual.** Added a case to `authentication.canonical-signer.spec.ts` signed with the current 6.x payload (method/path/timestamp lowercased, metadata joined *verbatim*) — the existing `signedRequest` helper folds the whole string, which would have made a re-cased key fail on the signature and pass for the wrong reason on every version. A positive control on the same new helper pins that it produces genuinely valid signatures.

Verified counterfactually rather than assumed — pinned back to 6.2.0, the new case fails:

```
● and the scene signer is delivered under a re-cased key
  › should reject it rather than read the metadata as carrying no signer

  expect(received).rejects.toThrow()
  Received promise resolved instead of rejected
  Resolved to value: "0xe1ef...7900"
```

Restored to 6.3.0: 17/17 pass. The positive control passes on both versions, so the failure is attributable to the gate and not to a broken signature. `tsc --noEmit` clean; `tslint` reports the same 5 pre-existing prettier findings as the untouched file at `HEAD` (identical, shifted by the length of the addition) — none introduced here.

